PR #3, opened 2021-11-03
Ar Update yaml loader

gokhanalcan:
In pyyaml 6.0, yaml.load() function gives the following error:

> TypeError: load() missing 1 required positional argument: 'Loader'

Either reverting back to pyyaml==5.4.1 or using yaml.safe_load() <[ref](https://stackoverflow.com/questions/1773805/how-can-i-parse-a-yaml-file-in-python/1774043#1774043)> solve the problem.


JacopoPan:
Hi @gokhanalcan,

the solution we have in the `main` branch is to use `yaml.load(f, Loader=yaml.FullLoader)`.
I'll have a look at what makes the most sense in the `ar` branch.

PR #9, opened 2022-02-07
Add poetry pyproject toml

amrit110:
Add pyproject.toml file which contains dependencies
Update README to include instructions to setup environment using venv, and install dependencies using poetry
Update python version to be 3.8.10 (tested experiment scripts)


amrit110:
Hi @JacopoPan , I've added a pyproject.toml file and also instructions to install using poetry which is sort of a nice alternative install method to using setup.py. Furthermore, I also recommend freezing dependency versions in setup.py in the future (I can add this too based on what you think). I also think that using python 3.8 for the repo might be a good idea since it ships as default python3 with ubuntu 20.04, unless you feel otherwise. I did run the examples and experiments to see if the update breaks anything and I don't see any issues. In the future, any changes to python version or dependencies should be tested, and we can work towards that.

JacopoPan:
Thanks! I agree with you your choice of 3.8 for the new instructions.
I’m agnostic about freezing dependencies. I think it’s our role as maintainers to keep the code properly working with the most recent external packages (so either not freezing them or frequently updating them).

PR #14, opened 2022-03-15
Refactor/device parameter

catgloss:
Changes:

For simplicity removed the device parameters and made it so that the command line --use_gpu could be used across all controllers
Added the parameter to BaseController class and gp_mpc which doesn't inherit BaseController.
Modified configuration.py to include the use_gpu argument as well and add it to the config dict
Notes:
-This change was mostly made for ease of use with gpu - it prevents any issues with the ambiguity of device. Note that use_gpu must be based in the make function for it to be included


JacopoPan (commented):
LGTM, @adamhall @Justin-Yuan plz confirm

adamhall (approved):
Tested code out for GPMPC and all checks out.

PR #15, opened 2022-03-15
Inertial_prop parameter and init typechecking 

catgloss:
Changes:

Added checks for dimension and type of inertial prop in cartpole, gp_mpc. This prevents an error being thrown when you try to use gp_mpc and cartpole together since they share the inertial_prop variable.
Also added some typechecking in the inits for the controller classe ppo and base_controller.
Notes:

I can extend the type checking to all the other controllers if wanted


JacopoPan (changes requested):
Thanks, good point but for the sake of reusability, I wouldn't hardcode inertial properties dimensions in the GP-MPC controller. @adamhall will update the controller to automatically cope with different `inertial_prop`

catgloss:
> Thanks, good point but for the sake of reusability, I wouldn't hardcode inertial properties dimensions in the GP-MPC controller. @adamhall will update the controller to automatically cope with different `inertial_prop`

that makes sense, thank you for the review

PR #16, opened 2022-03-15
Fixing failure when plotting with gp_mpc

catgloss:
Changes:

Added a check for agent type in plot for gp_mpc. This prevents the code from failing when plotting is set to true when doing a cartpole and gp_mpc combination
Added plot parameter to the config file so that code doesn't need to be directly modified to run with plotting (previously was an input to learn()

Notes:
I'm not sure if it was omitted on purpose, but at least the error doesn't prevent it from running anymore. I also put an arbitrary initial state in place for plotting but it may need to be modified. The algorithm can solve with this initial state but it is sensitive to large deviations from 0 in x, x dot, theta, theta dot.


JacopoPan (commented):
@adamhall will remove the plotting method that is not meant for use beyond his own debugging

PR #19, opened 2022-03-24
Fix dependencies

catgloss:
In pyproject.toml: 
- fixing the python version to be 3.8.x to avoid issues on the cluster 
- fixing casadi to be version 3.5.5 
- fixing pycddlib to be version 2.1.4
- fixing pybullet to be version 3.2.0

Additionally, adding note to README.md about installing gmp, a dependency of pycddlib, that the repo won't build without. I've also included the poetry.lock file so that poetry update can be run instead of poetry install


JacopoPan (commented):
Should we consider adding the .lock file to .gitignore?

Is there are reason NOT to allow more recent subversion of cased/pybullet?

catgloss:
> Should we consider adding the .lock file to .gitignore?
> 
> Is there are reason NOT to allow more recent subversion of cased/pybullet?

I had some issues when building with the more recent versions, so i fixed it just to ensure no issues when building. Feel free to try on your own system.

JacopoPan:
I added a note in README.md on the installation of `gmp` with `conda` or `apt-get`
For now, I would not make the dependencies in `pyproject.toml` more restrictive

PR #20, opened 2022-03-27
Constructing limit array in gp_mpc

catgloss:
When limits are used to create samples, they are read in the order of insertion into the dictionary. This does not necessarily match the order of the definition of the init state (which can be an array or dict). 
To ensure consistency, manually define the order of the limits in relation to the parameter they correspond to. This is not a necessary change but just makes it more consistent and less prone to error from user defined info.


JacopoPan (commented):
Also a good point but I think we should "just" add a way to automatically check dictionary with a keys ordering predefined from within an environment (so that it can be re-used across different dynamical systems)

JacopoPan:
Thanks, will be addressed when we merge the next version of `GP-MPC` (from @adamhall)

PR #22, opened 2022-04-07
Make slack parameter into list if initialized as int/float 

catgloss:
Change:
Making slack into list of length obs dim when not already a list to ensure no failure at line 168 where slack is indexed. 

Additional Comments: 
Might be good to have a length check on if it is a list as well


JacopoPan (approved):
LGTM and approved by @Justin-Yuan

PR #27, opened 2022-05-03
Adding BaseSafetyFilter Class and MPSCs as Safety Filters

Federico-PizarroBejarano:
Currently MPSCs and CBFs are implemented as controllers. Instead this adds safety filters as their class that can be passed to controllers to certify their output, and can be specified on the command-line similar to controllers. MPSCs are re-implemented using this new class and various experiments are laid out for cartpole and quadrotor stabilization. The MPSC has been improved slightly. The PR for CBFs will build on top of this one.


Federico-PizarroBejarano:
One decision remains which is whether to remove the current implementation of MPSCs and CBFs. This would break the current annual_reviews figures. They could be redone with the current implementation, but will almost certainly be slightly different as the implementation is different, although equally or more correct since there have been several updates. Since the code will continue changing, staying tied to reproducing the original figures seems counter-productive, so I think the best approach would be to have different versions of the gym that can reproduce the figures of different papers, but continue updating it and changing code.

Federico-PizarroBejarano:
Adding safety filters in an experiment requires `from safe_control_gym import safety_filters` to be added to the experiment file. I am not sure why. Any ideas?

Federico-PizarroBejarano:
Fixed both issues above. Figure 8 has been replicated, and I wasn't importing all the safety filters in one of the inits.

Federico-PizarroBejarano:
Going to break this up into smaller PRs

PR #31, opened 2022-05-26
Bugfix/ilqr loop

catglossop:
Changes: 
Fixed functionality for getting frame results output 

Comments: 
Due to looping, frames need to be added for each set of k steps (in one iteration) as blocks and the full list needs to be extended. Previously, just one frame was extracted due to this issue.

### DIFF
--- a/safe_control_gym/controllers/lqr/ilqr.py
+++ b/safe_control_gym/controllers/lqr/ilqr.py
@@ -174,176 +174,179 @@ class iLQR(BaseController):
         # Set update unstable flag to False
         self.update_unstable = False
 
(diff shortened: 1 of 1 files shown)


JacopoPan:
@catglossop 

I believe you have used this for the cartpole experiments.
Have you tested it with the quadrotor as well?

catglossop:
> @catglossop
> 
> I believe you have used this for the cartpole experiments. Have you tested it with the quadrotor as well?



> 

Yes, I tested the functionality with the quadrotor as well and it is functional.

JacopoPan (approved):
LGTM

PR #33, opened 2022-06-04
bug fix for explicit dynamics of 2d quadrotor

hannakrasowski:
If the dynamics are changed to the explicit dynamics for the tracking example (tracking.yaml, line 7 -> physics: dyn) and tracking.py is run, the following error occurs:

**Commands**
```
cd examples
python3 tracking.py --overrides ./tracking.yaml
```

**Error stack**
```
Traceback (most recent call last):
  File "tracking.py", line 77, in <module>
    main()
  File "tracking.py", line 58, in main
    results = ctrl.run(iterations=ITERATIONS)
  File "/home/hanna/Data/Research/SafeRL/safe-control-gym/safe_control_gym/controllers/pid/pid.py", line 97, in run
    obs, reward, done, info = self.env.step(action)
  File "/home/hanna/Data/Research/SafeRL/safe-control-gym/safe_control_gym/envs/gym_pybullet_drones/quadrotor.py", line 367, in step
    super()._advance_simulation(rpm, disturb_force)
  File "/home/hanna/Data/Research/SafeRL/safe-control-gym/safe_control_gym/envs/gym_pybullet_drones/base_aviary.py", line 285, in _advance_simulation
    self._dynamics(clipped_action[i, :], i)
  File "/home/hanna/Data/Research/SafeRL/safe-control-gym/safe_control_gym/envs/gym_pybullet_drones/base_aviary.py", line 552, in _dynamics
    vel = vel + self.TIMESTEP * no_pybullet_dyn_accs
AttributeError: 'Quadrotor' object has no attribute 'TIMESTEP'
```

This is because the attribute self.TIMESTEP is not defined. Minor changes in the _dynamics function solve this issue. A proposal how to fix this is created as pull request.

### DIFF
--- a/safe_control_gym/envs/gym_pybullet_drones/base_aviary.py
+++ b/safe_control_gym/envs/gym_pybullet_drones/base_aviary.py
@@ -549,10 +549,10 @@ class BaseAviary(BenchmarkEnv):
         rpy_rates_deriv = np.dot(self.J_INV, torques)
         no_pybullet_dyn_accs = force_world_frame / self.MASS
         # Update state.
-        vel = vel + self.TIMESTEP * no_pybullet_dyn_accs
-        rpy_rates = rpy_rates + self.TIMESTEP * rpy_rates_deriv
-        pos = pos + self.TIMESTEP * vel
-        rpy = rpy + self.TIMESTEP * rpy_rates
+        vel = vel + self.PYB_TIMESTEP * no_pybullet_dyn_accs
+        rpy_rates = rpy_rates + self.PYB_TIMESTEP * rpy_rates_deriv
+        pos = pos + self.PYB_TIMESTEP * vel
+        rpy = rpy + self.PYB_TIMESTEP * rpy_rates
         # Set PyBullet's state.
         p.resetBasePositionAndOrientation(self.DRONE_IDS[nth_drone],
                                           pos,


JacopoPan (changes requested):
Thanks for noting this @hannakrasowski 

I also now see that `self.TIMESTEP` is used in `def render(self, mode='human', close=False):` of `base_aviary`

I am not fully sure it's ideal to use the `PYB_` constant when Bullet is not used, I'd probably prefer something like:
```
def _reset_simulation(self):
    if self.PHYSICS == Physics.DYN:
        self.TIMESTEP = ...  # or self.DYN_TIMESTEP
```
Also, what is the reason for you to choose `Physics.DYN` over `Physics.PYB`?

JacopoPan:
I'm closing this PR because I think the usage should be limited to PyBullet-based physics updates and the CasADi symbolic model. I would instead consider merging a PR that entirely strips `BaseAviary` of `PHY.DYN`.

JakobThumm:
Thanks Jacobo, the DYN function helped us nevertheless to develop a safety method. 
Is there a documentation with the exact dynamics equations of the pybullet 2D quadrotor already?

JacopoPan:
@JakobThumm the 2D dynamics are in https://arxiv.org/pdf/2109.06325.pdf, or did you mean 3D?

JakobThumm:
I used the https://github.com/utiasDSL/safe-control-gym/blob/3e2d93cf4c9cbcd3facbc9ccf040511b51243674/safe_control_gym/envs/gym_pybullet_drones/base_aviary.py#L515 function to get the discrete system dynamics. But maybe this is trivial.

PR #35, opened 2022-06-07
Align thrust and pwm limits in quadrotor environment

### LINKED ISSUE #34: Align thrust and pwm limits in quadrotor environment
Hi,

in the quadrotor environment, is there any particular reason why the maximal thrust does not correspond to the maximum PWM?
If I understand it correctly, the thurst is converted into PWM and then into RPM.
The PWM is clipped to:
https://github.com/utiasDSL/safe-control-gym/blob/523d02df1a5616aeb6f666392dd2b5528f7332fa/safe_control_gym/envs/gym_pybullet_drones/quadrotor_utils.py#L47
Since in RL, the action output is usually also clipped
https://github.com/utiasDSL/safe-control-gym/blob/523d02df1a5616aeb6f666392dd2b5528f7332fa/safe_control_gym/envs/gym_pybullet_drones/quadrotor.py#L557
These two clippings should be the same.

Currently, the maximum thrust is defined as
https://github.com/utiasDSL/safe-control-gym/blob/523d02df1a5616aeb6f666392dd2b5528f7332fa/safe_control_gym/envs/gym_pybullet_drones/base_aviary.py#L141

I would suggest to change that to
```
self.MIN_THRUST = self.KF * 4 * (self.PWM2RPM_SCALE * self.MIN_PWM + self.PWM2RPM_CONST)**2
self.MAX_THRUST = self.KF * 4 * (self.PWM2RPM_SCALE * self.MAX_PWM + self.PWM2RPM_CONST)**2
```
in the constructor of `quadrotor.py` you would need to adjust that value for the 1D and 2D cases:
```
if self.QUAD_TYPE == QuadType.ONE_D:
    self.MAX_THRUST *= 1/4
elif self.QUAD_TYPE == QuadType.TWO_D:
    self.MAX_THRUST *= 1/2
```
EDIT: Added `**2` to formula.

JakobThumm:
This merge request fixes #34 .
The formula for min and max thrust input is taken from 
https://github.com/utiasDSL/safe-control-gym/blob/523d02df1a5616aeb6f666392dd2b5528f7332fa/safe_control_gym/envs/gym_pybullet_drones/quadrotor_utils.py#L40

### DIFF
--- a/safe_control_gym/envs/gym_pybullet_drones/quadrotor.py
+++ b/safe_control_gym/envs/gym_pybullet_drones/quadrotor.py
@@ -487,8 +487,10 @@ class Quadrotor(BaseAviary):
                                            dtype=np.float32)
         else:
             # direct thrust control 
-            self.action_space = spaces.Box(low=np.zeros(action_dim), 
-                                           high=self.MAX_THRUST * np.ones(action_dim), 
+            a_low = self.KF * action_dim * (self.PWM2RPM_SCALE * self.MIN_PWM + self.PWM2RPM_CONST)**2
+            a_high = self.KF * action_dim * (self.PWM2RPM_SCALE * self.MAX_PWM + self.PWM2RPM_CONST)**2
+            self.action_space = spaces.Box(low=np.full(action_dim, a_low, np.float32), 
+                                           high=np.full(action_dim, a_high, np.float32), 
                                            dtype=np.float32)
 
     def _set_observation_space(self):


JacopoPan (approved):
This seems a good idea to me (maybe it could be improved by making the action space and the action clipping in `_preprocessing` automatically align to one another).

Justin-Yuan (approved):

safe_control_gym/envs/gym_pybullet_drones/quadrotor.py
```diff
@@ -487,8 +487,10 @@ def _set_action_space(self):
                                            dtype=np.float32)
         else:
             # direct thrust control 
-            self.action_space = spaces.Box(low=np.zeros(action_dim), 
-                                           high=self.MAX_THRUST * np.ones(action_dim), 
+            a_low = self.KF * action_dim * (self.PWM2RPM_SCALE * self.MIN_PWM + self.PWM2RPM_CONST)**2
```
JacopoPan, line 490: @JakobThumm on a second look, I am not sure that multiplying by `action_dim` is entirely correct: it would work well for the 2D quadrotor but, in the 1D case we are essentially commanding all 4 motors at once, so we have a theoretical 4x max thrust (the way you wrote it would be multiplied by 1) and for the 3D quadrotor we will want to take 1x max thrust you compute on each motor (but multiplying by `action_dim` would make it 4 times larger)

JacopoPan (changes requested):

safe_control_gym/envs/gym_pybullet_drones/quadrotor.py
JacopoPan: Note that the previous, looser, possibly unachievable, upper bound was using a `self.MAX_THRUST` compute for all 4 motors collectively

JakobThumm:
You are right, it should use `n_motor` as in 
https://github.com/utiasDSL/safe-control-gym/blob/19ec4f9a6bf7a5c2691aee7c281b0ee255d2e7cc/safe_control_gym/envs/gym_pybullet_drones/quadrotor_utils.py#L38
instead of action_dim.
I corrected this.

JacopoPan (approved):

PR #36, opened 2022-06-18
Implement check and warning of at most 1 active constraint for safe explorer PPO 

Justin-Yuan:
The added code checks for the assumption of at most 1 active constraint violation, it throws a warning if the given batch does not satisfy the constraint, but does not terminate the program given the overall controller still learns to encourage constraint satisfaction. The added comment also suggests alternatives if the user needs strict constraint satisfaction.


JacopoPan (approved):
LGTM

PR #44, opened 2022-07-18
IPOPT not working for GPMPC on AR fig 6

adamhall:
IPOPT not working properly for GPMPC on the Annual Reviews Figure 6 experiment.


adamhall:
The issue was that IPOPT can slightly violte constraints (1e-9). For whatever reason, this was happening now. I've added a constraint_tol parameter to mpc, linear_mpc, and gp_mpc that shrinks the constraints by 1e-8 (default) to avoid this issue. Reran all GPMPC scripts and all seem to run okay and give results. Added new models as even this small constraint change seemed to have some impacts.

PR #45, opened 2022-07-18
Add DDPG baseline controller

Justin-Yuan:
The DDPG implementation is largely based on SAC's implementation. The notable addition is applying action noise during training. The action noise implementation is based on [Shangtong's repo](https://github.com/ShangtongZhang/DeepRL/blob/master/deep_rl/agent/DDPG_agent.py), with license header kept in the scripts under `safe_control_gym/math_and_models/random_process.py` and `safe_control_gym/math_and_models/scheduler.py`. Action noise can be configured via the config, an example of using OrnsteinUhlenbeckProcess is shown in the default `ddpg.yaml`


Justin-Yuan:
The requested changes 1-4 are pushed to commit 09de07a884e715948c7ae644f3216ff08280d5bf. I think `schedule.py` is of general use and can be used for learning rates for example, hence put it under `math_and_modules/`.

JacopoPan (approved):

JacopoPan (approved):

Issue #42, opened 2022-07-13
Re-create Figure 6 in "Safe Learning in Robotics" errors


jwangjie:
Hey Jacopo,

Thanks for the great review paper and open sourcing the accompanying codes here. 

I would like to reproduce Figure 6 by following [commands](https://github.com/utiasDSL/safe-control-gym#figure-6robust-gp-mpc-1) but got error messages after successfully loading five GP models as
```
#########################################
#       Loading GP dimension 5         #
#########################################
Path: ./trained_gp_model/best_model_5.pth
Loaded!
Traceback (most recent call last):
  File "./gp_mpc_experiment.py", line 179, in <module>
    ctrl.learn()
  File "/home/jwang/Control_ws/src/safe-control-gym/safe_control_gym/controllers/mpc/gp_mpc.py", line 809, in learn
    self.reset()
  File "/home/jwang/Control_ws/src/safe-control-gym/safe_control_gym/controllers/mpc/gp_mpc.py", line 899, in reset
    self.setup_gp_optimizer(n_ind_points)
  File "/home/jwang/Control_ws/src/safe-control-gym/safe_control_gym/controllers/mpc/gp_mpc.py", line 562, in setup_gp_optimizer
    mean_post_factor_val, Sigma, K_zind_zind_inv, z_ind_val = self.precompute_sparse_gp_values(n_ind_points)
  File "/home/jwang/Control_ws/src/safe-control-gym/safe_control_gym/controllers/mpc/gp_mpc.py", line 394, in precompute_sparse_gp_values
    inds, dist_mat = pairwise_distances_argmin_min(centroids, inputs[:, self.input_mask])
  File "/home/jwang/anaconda3/envs/safe/lib/python3.8/site-packages/sklearn/metrics/pairwise.py", line 680, in pairwise_distances_argmin_min
    values, indices = PairwiseDistancesArgKmin.compute(
  File "sklearn/metrics/_pairwise_distances_reduction.pyx", line 672, in sklearn.metrics._pairwise_distances_reduction.PairwiseDistancesArgKmin.compute
  File "sklearn/metrics/_pairwise_distances_reduction.pyx", line 1055, in sklearn.metrics._pairwise_distances_reduction.FastEuclideanPairwiseDistancesArgKmin.__init__
  File "sklearn/metrics/_dist_metrics.pyx", line 1300, in sklearn.metrics._dist_metrics.DatasetsPair.get_for
[...]

PR #46, opened 2022-07-18
Changing Dependence on sklearn.PairwiseDistancesArgKmin

adamhall:
As per issue #42, using the function `PairwiseDistancesArgKmin` from `sklearn` doesn't seem to work for the latest versions (1.1.1). We will either change the function/modify the code to get around this dependency, OR modify the installation to force `sklearn` to an older version.


adamhall:
The issue is that the new version of `pairwise_distances_argmin_min` requires c-contiguous arrays. Numpy arrays can be c-contiguous or f-contiguous (fortran) and [it has to do with how they are stored in memory](https://stackoverflow.com/questions/26998223/what-is-the-difference-between-contiguous-and-non-contiguous-arrays). `inputs` itself is c-contiguous, but when the slice `inputs[:, self.input_mask]` is taken, it becomes f-contiguous. Thus, I added a single line to make sure the slice that is taken is c-contiguous. This seems to work with the default version of `scikit-learn`.

JacopoPan (approved):
LGTM

PR #52, opened 2022-07-26
Adding obs_goal_horizon to cartpole and fixing reward

Federico-PizarroBejarano:
All changes are non-breaking EXCEPT making obs_goal_horizon default to 1 for quadrotor. I can switch this back but I found it unintuitive when I started. Since I didn't know about obs_goal_horizon, I could not figure out why the observations were twice as large as they should have been and had to ask Justin. I feel like the default behaviour should be not extending the observation and only extending it if asked.

### DIFF
--- a/safe_control_gym/envs/benchmark_env.py
+++ b/safe_control_gym/envs/benchmark_env.py
@@ -379,6 +379,22 @@ class BenchmarkEnv(gym.Env):
         processed_action = self._preprocess_control(action)
         return processed_action
     
+    def extend_obs(self, obs, next_step):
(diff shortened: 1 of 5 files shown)


Justin-Yuan:
LGTM with 1 small thing, can you also add the`obs_goal_horizon` arg to the template `cartpole.yaml` as well?

JacopoPan (changes requested):
Looks good but please address @Justin-Yuan 's comment in the conversation section before I merge

Federico-PizarroBejarano:
Done, tell me how that looks!

Justin-Yuan (approved):

JacopoPan (approved):

PR #56, opened 2022-07-28
Updating the experiment class and related files

Federico-PizarroBejarano:
A PR with just the changes to the experiment class. Once we agree and merge this, I will PR the PID changes.

### DIFF
--- a/safe_control_gym/experiment.py
+++ b/safe_control_gym/experiment.py
@@ -1,148 +1,238 @@
-"""To standardize training/evaluation interface.
+"""To standardize training/evaluation interface. """
 
-Todo:
-    * 
-    
-"""
-import os 
 import gym
 import numpy as np 
 from copy import deepcopy
-import functools
-from collections import defaultdict, deque
+from collections import defaultdict
 from termcolor import colored
 
-# from safe_control_gym.utils.configuration import ConfigFactory
-# from safe_control_gym.utils.registration import make
-# from safe_control_gym.utils.utils import mkdirs, set_dir_from_config, set_device_from_config, set_seed_from_config, save_video
-from safe_control_gym.utils.utils import get_random_state, set_random_state, is_wrapped
+from safe_control_gym.utils.utils import is_wrapped
 from safe_control_gym.math_and_models.metrics import compute_cvar
 
 
-
 class Experiment:
+    """Generic Experiment Class. """
+
+    def __init__(self, 
+                 env, 
+                 ctrl, 
+                 safety_filter = None,
+                ):
+        """Creates a generic experiment class to run evaluations and collect standard metrics.
+
+        Args:
+            env (gym.Env): the environment for the task.
+            ctrl (BaseController): the controller for the task
+            safety_filter (BaseSafetyFilter): the safety filter to filter the controller
+        """
+        
+        self.env = env
+        if not is_wrapped(self.env, RecordDataWrapper):
+            self.env = RecordDataWrapper(self.env)
+        self.ctrl = ctrl
+        self.safety_filter = safety_filter
     
-    def __init__(self):
-        # what to put here? 
-        # now the class seems like a collection of utils (from `main.py`) only.
-        pass
+    def run_evaluation(self, n_episodes=10, log_freq=None, n_steps=None, **kwargs):
+        """Evaluate a trained controller. 
+        
+        Args:
+            n_episodes (int): number of runs to execute
+            n_steps (int): the number of steps to collect in total
+            log_freq (int): the frequency with which to log information
+        
+        Returns:
+            trajs_data (defaultdict(list)): the raw data from the executed runs 
+            metrics (dict): the metrics calculated from the raw data
+        """
+
+        self.reset()
+        trajs_data = self._execute_evaluations(log_freq=log_freq, n_episodes=n_episodes, n_steps=n_steps, **kwargs)
+        metrics = self.compute_metrics(trajs_data)
+        
+        # terminal printouts
+        for metric_key, metric_val in metrics.items():
+            print("{}: {:.3f}".format(colored(metric_key,"yellow"), metric_val))
+        print("Evaluation done.")
+        return trajs_data, metrics
     
-    def launch_training(self, config, ctrl):
+    def _execute_evaluations(self, n_episodes=10, n_steps=None, log_freq=None, **kwargs):
+        """Runs the experiments and collects all the required data.
+
+        Args:     
+            n_episodes (int): number of runs to execute
+            n_steps (int): the number of steps to collect in total
+            log_freq (int): the frequency with which to log information
+        
+        Returns:
+            trajs_data (defaultdict(list)): the raw data from the executed runs
+        """
+
+        if n_episodes is None and n_steps is None:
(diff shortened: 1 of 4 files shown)


safe_control_gym/experiment.py
```diff
@@ -1,148 +1,238 @@
...
+        elif n_steps is not None:
+            while steps < n_steps: 
+                action = self.ctrl.select_action(obs, info)
+                # inner sim loop to accomodate different control frequencies
+                for _ in range(sim_steps):
+                    obs, reward, done, info = self.env.step(action)
+                    steps += 1 
+                    if steps >= n_steps:
+                        self.env.save_data()
+                        for data_key, data_val in self.ctrl.results_dict:
```
Justin-Yuan, line 106: it should be `self.ctrl.results_dict.items()` here? i think there's another occurrence of the somewhere.

safe_control_gym/experiment.py
```diff
@@ -1,148 +1,238 @@
...
             print("{}: {:.3f}".format(colored(metric_key,"yellow"), metric_val))
         print("Evaluation done.")
-        return results
+        return trajs_data, metrics
     
-    def collect_trajs(self, ctrl, env, log_freq=None, n_trajs=None, n_steps=None, **kwargs):
-        """"""
-        if not is_wrapped(env, RecordDataWrapper):
-            env = RecordDataWrapper(env)
+    def _execute_evaluations(self, n_episodes=10, n_steps=None, log_freq=None, **kwargs):
```
Justin-Yuan, line 58: with the two error checks, the `n_episodes` arg default should be `None` ?

safe_control_gym/experiment.py
```diff
@@ -1,148 +1,238 @@
...
     def __init__(self, env, deque_size=None, **kwargs):
         super().__init__(env)
-        self.deque_size = deque_size
-        self.initialize_data_containers()
-
-    def initialize_data_containers(self):
-        """"""
         self.episode_data = defaultdict(list)
-        self.data = defaultdict(lambda: deque(self.deque_size))
+        self.data = defaultdict(list)
```
Justin-Yuan, line 223: food for thought here: the reason for using `deque` is to consider logging only the latest `deque_size` number of episode data during training, otherwise we might be caching old data as well (unless there's also reset of `env.data` after each metric evaluation call).

safe_control_gym/experiment.py
```diff
@@ -1,148 +1,238 @@
...
         # initialize
-        sim_steps = log_freq // env.CTRL_FREQ if log_freq else 1 
-        n_step, n_traj = 0, 0
-        done = False
+        sim_steps = log_freq // self.env.CTRL_FREQ if log_freq else 1 
+        steps, trajs = 0, 0
         ctrl_data = defaultdict(list)
-        if env.INFO_IN_RESET:
-            obs, info = env.reset()
+        if self.env.INFO_IN_RESET:
```
Justin-Yuan, line 79: given the reset of each test episode is abstracted to `_evaluation_reset()`, i think it's also possible to use it here, but provide `ctrl_data=None` to indicate nothing is lingering from the past episode.  correspondingly, add logic to deal with empty `ctrl_data` in `_evaluation_reset()`

Justin-Yuan (changes requested):
nice overall. besides the specific comments, i also have some general questions
- given the current `__init__` signature, how would you deal with separate training env and eval env? (since the two have different configs)   
- it also seems like we are supposed to call `run_evaluation()` to get the data metrics? but since we are likely to call it periodically during training, it might conflict with having a `self.reset()` call in `run_evaluation()`   (specifically the `self.ctrl.reset()` call inside `self.reset()` that resets training globally). maybe the bigger question is how do you imagine `run_evaluation()` being used in the training loop?  (can reference [this](https://github.com/utiasDSL/safe-control-gym/blob/1179991344e49e7ac9174305c550b009ce2b1824/safe_control_gym/controllers/ppo/ppo.py#L177), how ppo currently does the periodic evaluation)
- there seems to be similar code on the data container for both `RecordDataWrapper` and for the `ctrl_data`, we could consider extracting a common data struct in `utils` then use it for both?   (essentially it might be a nested defaultdict of list with a bunch of helper methods)

safe_control_gym/experiment.py
Federico-PizarroBejarano: Hmm maybe its cleaner to have a reset function that resets env.data since I think its kind of weird that the data would just randomly get full. In experiment.reset() we could clear the env.data

safe_control_gym/experiment.py
Federico-PizarroBejarano: You're totally right, I'll fix that now

safe_control_gym/experiment.py
Federico-PizarroBejarano: Hmm I forgot to test this, I'll test it now

safe_control_gym/experiment.py
Federico-PizarroBejarano: That's smart, will do

Federico-PizarroBejarano:
> 

@Justin-Yuan thanks for all the comments, I'll try and fix them soon

Related to your comments:

1. I originally thought we could pass in a training_env to the __init__ and it would pass it to the ctrl.learn(). If its None, it would simply let the learn() choose its own interior train. Would that work?
2. Hmm I haven't considered that. I also think we should do things one at a time: make sure it works and is a good structure for PID and then move on to RL stuff. However, obviously we should think ahead and make sure its amenable to RL changes. We could pass in a flag to tell the evaluation that its doing a RL update, or we could not reset the controller at all i guess.
3. I am not sure what you mean by this, we can have a call or something next week or just talk on slack.

safe_control_gym/experiment.py
Justin-Yuan: if we don't want it to get full, can set `deque_size` to None so it's the same as a list. 

two following questions are 
- is `experiment.reset()` meant to be called once per training or many times before each `run_evaluation()` call?  (related to the general question #2 at the bottom) 
- are we trying to log data from the training env as well? and if so how?    (this is coz the wrapper and `deque` were originally designed to log training env data, say ppo wants to check the latest 10 training episodes' data or the current total number of constraint violations so far)

Federico-PizarroBejarano: I added a flag in run_evaluation that does not reset the controller if it is training. Also added a clear_data function to the environment that can be called when reseting the entire experiment. I have left it as a list since idk the benefit of a deque if it has no size and we are only appending to the end, but if you think its better we can switch to deque. I don't think this solves exactly how we will do RL training, but I think it addresses the immediate comments

JacopoPan (commented):
This PR merges into `dev-experiment-class` so I am ok with it as long as @Justin-Yuan approves it. I will have further comments when we merge `dev-experiments-class` into `main`

safe_control_gym/experiment.py
Federico-PizarroBejarano: For RL, need top 10 items or something like that. Will change for RL

Federico-PizarroBejarano:
Change train_env in learn() to instead pass in a dict that overloads the config and use env_func() to build that training environment

PR #63, opened 2022-08-11
Fixing typos

Federico-PizarroBejarano:
Found some typos of various kinds

### DIFF
--- a/safe_control_gym/controllers/mpc/gp_mpc.py
+++ b/safe_control_gym/controllers/mpc/gp_mpc.py
@@ -152,6 +152,7 @@ class GPMPC(MPC):
             constraint_tol=constraint_tol,
             # runner args
             # shared/base args
+            use_gpu=use_gpu,
             output_dir=output_dir,
             additional_constraints=additional_constraints,
             **kwargs)
@@ -175,7 +176,6 @@ class GPMPC(MPC):
         self.learning_rate = learning_rate
         self.gp_model_path = gp_model_path
         self.normalize_training_data = normalize_training_data
-        self.use_gpu = use_gpu and torch.cuda.is_available()
         self.seed = seed
         self.prob = prob
         self.sparse_gp = sparse_gp
--- a/safe_control_gym/controllers/mpc/gp_mpc.py
+++ b/safe_control_gym/controllers/mpc/gp_mpc.py
@@ -152,12 +152,12 @@ class GPMPC(MPC):
             constraint_tol=constraint_tol,
             # runner args
             # shared/base args
-            use_gpu=use_gpu,
             output_dir=output_dir,
             additional_constraints=additional_constraints,
+            use_gpu=use_gpu,
+            seed=seed,
             **kwargs)
         # Setup environments.
-        self.env_func = env_func
         self.env = env_func(randomized_init=False, seed=seed)
         self.env_training = env_func(randomized_init=True, seed=seed)
         # No training data accumulated yet so keep the dynamics function as linear prior.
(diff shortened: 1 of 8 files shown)


safe_control_gym/controllers/mpc/gp_mpc.py
```diff
@@ -152,6 +152,7 @@ def __init__(
             constraint_tol=constraint_tol,
             # runner args
             # shared/base args
+            use_gpu=use_gpu,
```
JacopoPan, line 155: GPMPC parent class is MPC whose parent is BaseController
But MPC's constructor never calls super's constructor.
Is this correct?

safe_control_gym/envs/benchmark_env.py
```diff
@@ -386,6 +386,7 @@ def before_step(self, action):
         # Sanity check (reset at least once).
         self._check_initial_reset()
         # Save the raw input action.
+        action = np.array([action]).flatten()
```
JacopoPan, line 389: Do we want to forbid non-flat actions spaces?
Wouldn't a warning be more appropriate?

JacopoPan (changes requested):

safe_control_gym/envs/benchmark_env.py
Federico-PizarroBejarano: I found this error b/c some controllers return floats for the cartpole rather than a numpy array, which causes an error downstream. im guessing a controller could reasonably also send a 2D array. this extra line makes sure the action from this point forward is a 1D array

safe_control_gym/controllers/mpc/gp_mpc.py
Federico-PizarroBejarano: Damn you're right my bad, should I remove this or make MPC call the BaseController? I feel like every controller should call the BaseController super since it has some logic rather than repeating the logic, which will be more important if we ever decide to add more logic to BaseController as well

safe_control_gym/envs/benchmark_env.py
JacopoPan: I support making `action` a NumPy array, but I'd replace the flattening with a check on `action.ndim` that raises an exception so that the developer of a new controller can decide whether that was intentional or not

safe_control_gym/controllers/mpc/gp_mpc.py
JacopoPan: I think MPC should call the BaseController, although on the short run now (as @adamhall designed it as a base class itself) it might just bloat the object.

safe_control_gym/envs/benchmark_env.py
Federico-PizarroBejarano: Hmm but then we'd need a few if statements, something like this: 

`if type == float or int -> np.array([action])`
`if type == list -> np.array(action)`
`if action.ndim != 1 -> raise Exception`

How does that look?

Federico-PizarroBejarano: admittedly idk why it would be an int lol

safe_control_gym/controllers/mpc/gp_mpc.py
Federico-PizarroBejarano: wait so the conclusion is to add the call to BaseController to MPC even tho it bloats it a bit right?

safe_control_gym/envs/benchmark_env.py
JacopoPan: I think you could do `np.array([action]).squeeze()` instead of the first 2 lines, no?

safe_control_gym/controllers/mpc/gp_mpc.py
JacopoPan: yes with CC to @adamhall so that he can work on removing potentially duplicated code

safe_control_gym/envs/benchmark_env.py
Federico-PizarroBejarano: Yeah you right, I'll fix that now

JacopoPan (approved):

PR #66, opened 2022-08-12
Updating all controllers

Federico-PizarroBejarano:
Extracted some changes I had make for my own code, adding the `select_action` function to every controller. Few other small changes, plus cutting down LQR and iLQR significantly. We can discuss LQR and iLQR more but they were ~~essentially wrong~~ **outdated** before and it seems only I am using LQR. If we want to make sure everything is working I can create experiments for LQR and iLQR and add them as unit tests as well. 

Also moved some stuff around in MPC since I will need that function in the MPSC and don't want to have the same code in two places. Honestly reset_constraints could even be a more generic util than just an mpc util.


safe_control_gym/controllers/base_controller.py
```diff
@@ -54,6 +60,25 @@ def select_action(self, obs, info=None):
...
+        Args:
+            info (list): the info list returned from the environment.
+        
+        Returns:
+            step (int): the current step/iteration of the environment.
+        """
+        
+        if info is not None:
+            step = info['current_step']
+            if self.env.TASK == Task.TRAJ_TRACKING:
```
Federico-PizarroBejarano, line 75: Why is step too large

JacopoPan (commented):
This PR involves LQR/iLQR (@siqizhou, @Federico-PizarroBejarano), GP-MPC (@adamhall) and PPO/RARL/SAC (@Justin-Yuan), I'll merge as soon as all have left an approving review.

siqizhou (commented):
To summarize, the main changes to LQR/iLQR are (1) interface with env, (2) removing random initialization feature from the controller side (how is this done now? thru the experiment class?), and (3) removing plotting/print out for debugging. The core action selection functions and the iLQR updates are the same? Verify if the code is producing the same results/numbers as before to make sure the reset part you have is correct and also the updates at the end of each iteration are correct - it would be convenient to keep the printout for now for checking purposes unless there is another utils function to do this. Some of the printouts such as the error etc are useful for tuning purposes (e.g., the lambda parameter). There is a flag to turn off the plotting/debugging printouts and should not affect speed but the code can be reorganized.

adamhall (approved):
MPC and GPMPC look good.

PR #68, opened 2022-08-15
Linting dev-experiment Files

Federico-PizarroBejarano:
I installed pylint and pylance and linted the BaseController, Experiment, and the examples and pid experiment, plus the tests and the other files we've changed significantly. There should be no functional changes in this PR. All my future PRs will follow this style. Let me know if you agree with this general linting structure. 

Since these files either don't exist in the main or are being changes significantly in dev-experiment, I think it makes sense to lint them now. 

I also greatly improved the documentation. I think all the documentation in the files should be complete and follow a specific structure now.


safe_control_gym/controllers/base_controller.py
```diff
@@ -1,112 +1,106 @@
...
+        '''
+        return
 
     def reset(self):
-        """Do initializations for training or evaluation.
-
-        """
-        pass
+        '''Do initializations for training or evaluation. '''
+        return
```
JacopoPan, line 68: shouldn't this be something like 
```
raise NotImplementedError
```
if we want to make re-implementation mandatory?

safe_control_gym/envs/gym_pybullet_drones/quadrotor.py
```diff
@@ -1,11 +1,12 @@
-"""1D and 2D quadrotor environment using PyBullet physics.
+'''1D and 2D quadrotor environment using PyBullet physics.
```
JacopoPan, line 1: It's 3D too now, we should update this and below

JacopoPan (changes requested):
Just a couple of comments.
Other than that I scrolled through changes (and I also take your word that) it's all style changes and no functional changes.

safe_control_gym/controllers/base_controller.py
Federico-PizarroBejarano: You're right, fixed it. Also made them abstract methods in the other PR.

safe_control_gym/envs/gym_pybullet_drones/quadrotor.py
Federico-PizarroBejarano: Done

PR #71, opened 2022-08-17
Add similarity metrics: MSE(ED), DTW, LCSS, EDR, FD

Justin-Yuan:
Given these implemented similarity metrics, there are things to I notice before going forward 
1. they might be sensitive to the distance function we use to compare each pair of transition tuple
2. they might not need to use action sequences (only state sequences) 
3. the input trajectories might need to be normalized, and each element in the trajectory might need to be weighted/masked (on state dimensions we care about)
4. for now, they only compare two input trajectories, we need an outer loop to go through multiple pairs of such trajectories and get the average metric at the end. 
5. their compute time might be large (since each DTW, LCSS, EDR and FD of two trajs requires O(n^2), n is traj length)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,8 @@ experiments/pid/*data/
 results/
 z_docstring.py
 TODOs.md
-
+experiments/metrics/archived/
+experiments/metrics/temp_data/
 
 
 
--- a/experiments/metrics/config_overrides/cartpole/base.yaml
+++ b/experiments/metrics/config_overrides/cartpole/base.yaml
@@ -1,0 +1,42 @@
+task_config:
+  ctrl_freq: 50
+  pyb_freq: 1000
+  episode_len_sec: 5
+  normalized_rl_action_space: True
+  info_in_reset: True
+
+  # task 
+  task: stabilization
+  task_info:
+    stabilization_goal: [0]
+    stabilization_goal_tolerance: 0.005
+  cost: rl_reward
+
+  # init 
+  randomized_init: True
+  init_state_randomization_info:
+    init_x:
+      distrib: uniform
+      low: -0.3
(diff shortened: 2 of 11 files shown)


JacopoPan (changes requested):
@Justin-Yuan @adamhall please factor out something akin to a `distance(state1, state2, kwargs)` methods used by all metrics that we can successively re-implement

adamhall:
As we talked about, we should consider how to properly address the Metrics on SO(3). [This paper](https://www.cs.cmu.edu/~cga/dynopt/readings/Rmetric.pdf) gives a nice overview of the different metrics. I would suggest either metric 5 "deviation from the Identity matrix" (Section 3.5), or metric 6 "Geodesic on the Unit Sphere" (section 3.6) because we already have the machinery to compute the rotation matrices from the euler angles. The paper does state the metric 5 is the most computationally expensive, but I'm not sure how much this matters? It also states that metric 6 is the most geometrically meaningful.

Justin-Yuan:
As of commit ace4f7ab6ff0143b5c90e9a83191c3b297351d1b, the similarity metrics DTW, LCSS, EDR, FD all support a custom distance function between two states. A `state_distance()` function is added that accounts for proper rotation metric in state comparison. Other things to note about 
1. MMD has not adapted the custom distance function, since it involves using a kernel that already has an implicitly defined feature mapping for states
2. the `state_distance()` function uses Euclidean distance for angular velocities which might not be proper, @adamhall  is looking into this issue 
3. by adapting the custom state-pair distance in the trajectory similarity metrics, they might require further changes or investigations (marked as TODO lines), for example, substituting discrete 0/1 state-pair distance to the continuous distance value

JacopoPan (approved):

PR #77, opened 2022-08-27
Allow varying env model priors for PID, LQR, and MPC 

Justin-Yuan:
This PR is meant for 
1. current PID for quadrotor uses the overridden system params from env instead of the prior params, which should be fixed
2. Cartpole seems to have an argument `prior_prop` to set the prior inertial properties (also used in the benchmark experiment for LQR^150%), but Quadrotor does not have such an argument, we should add this missing functionality in Quadrotor.
3. currently the prior params in env are the same as the default URDF params, but we need to vary the priors for testing the similarity metrics

### DIFF
--- a/experiments/pid/config_overrides/quadrotor_stabilization.yaml
+++ b/experiments/pid/config_overrides/quadrotor_stabilization.yaml
@@ -43,3 +43,9 @@ task_config:
   episode_len_sec: 6
   cost: quadratic
   done_on_out_of_bound: False
+
+algo_config:
+  prior_info:
+    prior_prop:
+      M: 0.027
+      Iyy: 0.000014
--- a/experiments/pid/config_overrides/quadrotor_stabilization.yaml
+++ b/experiments/pid/config_overrides/quadrotor_stabilization.yaml
@@ -43,9 +43,3 @@ task_config:
   episode_len_sec: 6
   cost: quadratic
   done_on_out_of_bound: False
-
-algo_config:
-  prior_info:
-    prior_prop:
-      M: 0.027
-      Iyy: 0.000014
--- a/safe_control_gym/controllers/base_controller.py
+++ b/safe_control_gym/controllers/base_controller.py
@@ -128,10 +128,50 @@ class BaseController(ABC):
         '''Setup the results dictionary to store run information. '''
         self.results_dict = {}
         
-    def setup_prior(self):
-        '''Changes the prior model fetched from `self.env` for the controller.
+    def get_prior(self, env, prior_info={}):
+        '''Fetch the prior model from the env for the controller.
+
+        Note there's a default env.symbolic when each each env is created.
+        To make a different prior model, do the following when initializing a ctrl::
+
+            self.env = env_func() 
+            self.model = self.get_prior(self.env)
+
+        Besides the env config `base.yaml` and ctrl config `mpc.yaml`, 
+        you can define an additional prior config `prior.yaml` that looks like:: 
+
+            algo_config:
+                prior_info:
+                    M: 0.03
+                    Iyy: 0.00003
+
+        and to ensure the resulting config.algo_config contains both the params
+        from ctrl config and prior config, chain them to the --overrides like::
+
+            python main.py --algo mpc --task quadrotor --overrides base.yaml mpc.yaml prior.yaml ...
+        
+        Also note we look for prior_info from the incoming function arg first, then the ctrl itself.
+        this allows changing the prior model during learning by calling::
+        
+            new_model = self.get_prior(same_env, new_prior_info)
+        
+        Alternatively, you can overwrite this method and use your own format for prior_info
+        to customize how you get/change the prior model for your controller.
+
+        Args:
+            env (BenchmarkEnv): the environment to fetch prior model from.
+            prior_info (dict): specifies the prior properties or other things to 
+                overwrite the default prior model in the env.
+            
+        Returns:
+            SymbolicModel: CasAdi prior model.
         '''
-        prior_info = getattr(self, "prior_info", None)
-        if prior_info is not None:
-            # TODO: it does not work currently, since `_setup_symbolic()` doesn't take any args
-            self.env._setup_symbolic(prior_info)
+        if not prior_info:
+            prior_info = getattr(self, "prior_info", {})
+        # Note we only reset the symbolic model when prior_info is nonempty
+        if prior_info:
+            prior_model = env._setup_symbolic(prior_info)
+        # Note this ensures the env can still access the prior model, 
+        # which is used to get quadratic costs in env.step()
+        prior_model = env.symbolic
+        return prior_model
--- a/safe_control_gym/controllers/base_controller.py
+++ b/safe_control_gym/controllers/base_controller.py
@@ -142,9 +142,12 @@ class BaseController(ABC):
 
             algo_config:
                 prior_info:
-                    M: 0.03
-                    Iyy: 0.00003
-
+                    prior_prop:
+                        M: 0.03
+                        Iyy: 0.00003
+                    randomize_prior_prop: False
+                    prior_prop_rand_info: {}
+                    
         and to ensure the resulting config.algo_config contains both the params
         from ctrl config and prior config, chain them to the --overrides like::
 
@@ -168,9 +171,22 @@ class BaseController(ABC):
         '''
         if not prior_info:
             prior_info = getattr(self, "prior_info", {})
-        # Note we only reset the symbolic model when prior_info is nonempty
-        if prior_info:
-            prior_model = env._setup_symbolic(prior_info)
+        prior_prop = prior_info.get("prior_prop", {})
+
+        # randomize prior prop, similar to randomizing the inertial_prop in BenchmarkEnv
+        # this can simulate the estimation errors in the prior model
+        randomize_prior_prop = prior_info.get("randomize_prior_prop", False)
+        prior_prop_rand_info = prior_info.get("prior_prop_rand_info", {})
+        if randomize_prior_prop and prior_prop_rand_info:
+            # check keys, this is due to the current implementation of BenchmarkEnv._randomize_values_by_info()
+            for k in prior_prop_rand_info:
+                assert k in prior_prop, "A prior param to randomize does not have a base value in prior_prop."
+            prior_prop = env._randomize_values_by_info(prior_prop, prior_prop_rand_info)
+
+        # Note we only reset the symbolic model when prior_prop is nonempty
+        if prior_prop:
+            prior_model = env._setup_symbolic(prior_prop=prior_prop)
+
         # Note this ensures the env can still access the prior model, 
         # which is used to get quadratic costs in env.step()
         prior_model = env.symbolic
--- a/safe_control_gym/controllers/base_controller.py
+++ b/safe_control_gym/controllers/base_controller.py
@@ -127,3 +127,11 @@ class BaseController(ABC):
     def setup_results_dict(self):
         '''Setup the results dictionary to store run information. '''
         self.results_dict = {}
+        
+    def setup_prior(self):
+        '''Changes the prior model fetched from `self.env` for the controller.
+        '''
+        prior_info = getattr(self, "prior_info", None)
+        if prior_info is not None:
+            # TODO: it does not work currently, since `_setup_symbolic()` doesn't take any args
+            self.env._setup_symbolic(prior_info)
--- a/safe_control_gym/controllers/base_controller.py
+++ b/safe_control_gym/controllers/base_controller.py
@@ -185,7 +185,7 @@ class BaseController(ABC):
 
         # Note we only reset the symbolic model when prior_prop is nonempty
         if prior_prop:
-            prior_model = env._setup_symbolic(prior_prop=prior_prop)
+            env._setup_symbolic(prior_prop=prior_prop)
 
         # Note this ensures the env can still access the prior model, 
         # which is used to get quadratic costs in env.step()
--- a/safe_control_gym/controllers/pid/pid.py
+++ b/safe_control_gym/controllers/pid/pid.py
@@ -54,14 +54,12 @@ class PID(BaseController):
         super().__init__(env_func, **kwargs)
 
         self.env = env_func()
-        self.model = self.get_prior(self.env)
-
+        
         if self.env.NAME != Environment.QUADROTOR:
             raise NotImplementedError('[ERROR] PID not implemented for any system other than Quadrotor (2D and 3D).')
 
         self.env.reset()
         self.g = g
-        self.GRAVITY = g * self.model.quad_mass # The gravitational force (g*M) acting on each drone.
         self.KF = kf
         self.KM = km
         self.P_COEFF_FOR = np.array(p_coeff_for)
@@ -250,6 +248,8 @@ class PID(BaseController):
         '''Resets the control classes. The previous step's and integral
            errors for both position and attitude are set to zero.
         '''
+        self.model = self.get_prior(self.env)
+        self.GRAVITY = self.g * self.model.quad_mass # The gravitational force (g*M) acting on each drone.
         self.env.reset()
         self.reset_before_run()
 
--- a/safe_control_gym/controllers/pid/pid.py
+++ b/safe_control_gym/controllers/pid/pid.py
@@ -54,6 +54,7 @@ class PID(BaseController):
         super().__init__(env_func, **kwargs)
 
         self.env = env_func()
+        self.model = self.get_prior(self.env)
 
         if self.env.NAME != Environment.QUADROTOR:
             raise NotImplementedError('[ERROR] PID not implemented for any system other than Quadrotor (2D and 3D).')
@@ -61,7 +62,7 @@ class PID(BaseController):
         self.env.reset()
         self.g = g
         # self.GRAVITY = g * self.env.OVERRIDDEN_QUAD_MASS # The gravitational force (g*M) acting on each drone.
-        self.GRAVITY = g * self.env.PRIOR_MASS
+        self.GRAVITY = g * self.model.quad_mass
         self.KF = kf
         self.KM = km
         self.P_COEFF_FOR = np.array(p_coeff_for)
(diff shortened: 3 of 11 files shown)


Justin-Yuan:
There are two ways to do this
1. similar to what Cartpole currently has, we use an argument `prior_prop` when initializing the env, then the env's prior parameters will take the URDF params as default or overridden by `prior_prop` if it is non-empty. Then the prior params are used in `_setup_symbolic()` to define the prior model, and the model is used in the controller. an example config will be 
```
task_config:
    ...
    prior_prop:
        M: 0.03
```

2. alternatively, we can have a common method `setup_prior()` in `BaseController`, which is called after creating `self.env` but before fetching the symbolic model. The `setup_prior()` will call `_setup_symbolic(prior_info)` to re-define the symbolic model given the desired `prior_info`, which will now be an argument to the controller. Also, we need to change the `_setup_symbolic()` in each environment since it does not take in any arguments currently.  Then, an example config will be 
```
algo_config:
    ...
    prior_info:
        M: 0.03
```

approach 1 is easier coding-wise, but approach 2 is more intuitive since being able to change the prior should be part of the controller spec instead of env spec. For now, approach 1 is implemented in the commits as a temporary fix, but we should discuss about this.

Federico-PizarroBejarano:
Also @Justin-Yuan did you include a way to randomize the prior?

Justin-Yuan:
@JacopoPan @adamhall @Federico-PizarroBejarano , as of the latest commit, the abovementioned method 2 is implemented, including ways to get/change prior model in the controller. Additional features are also added:
- randomize prior properties
- assign X_EQ and U_EQ as attributes to the symbolic model so the controller can use them, remove them from the envs since they are not used anywhere

Justin-Yuan:
PID is modified to use the new prior format/interface. Specifically, the default prior will be the true inertial params used in the env (even if we initialize env with overriding `inertial_prop`), to use a different prior we need to add the prior specs in config yaml, for example in tracking,  
```
algo_config:
  prior_info:
    prior_prop:
      M: 0.027
      Iyy: 0.000014
```
(i)LQR and MPC controllers remain to be fixed as well.

experiments/pid/config_overrides/quadrotor_stabilization.yaml
```diff
@@ -43,3 +43,9 @@ task_config:
   episode_len_sec: 6
   cost: quadratic
   done_on_out_of_bound: False
+
+algo_config:
```
Federico-PizarroBejarano, line 47: could be cleaner to simply have a pid_config.yaml file with this instead of adding it in both quadrotor yamls

safe_control_gym/controllers/base_controller.py
```diff
@@ -127,3 +127,67 @@ def load(self,
...
+    def get_prior(self, env, prior_info={}):
+        '''Fetch the prior model from the env for the controller.
+
+        Note there's a default env.symbolic when each each env is created.
+        To make a different prior model, do the following when initializing a ctrl::
+
+            self.env = env_func() 
+            self.model = self.get_prior(self.env)
+
+        Besides the env config `base.yaml` and ctrl config `mpc.yaml`,
```
Federico-PizarroBejarano, line 140: why dont we just add the prior info to the controller yaml directly? and why is the defauly controller yaml `mpc.yaml`

adamhall:
As far as I can tell it looks okay. I will try and incorporate it with the mpc classes and see how it goes.

adamhall (approved):

safe_control_gym/controllers/pid/pid.py
```diff
@@ -54,13 +54,15 @@ def __init__(self,
...
         self.env = env_func()
+        self.model = self.get_prior(self.env)
 
         if self.env.NAME != Environment.QUADROTOR:
             raise NotImplementedError('[ERROR] PID not implemented for any system other than Quadrotor (2D and 3D).')
 
         self.env.reset()
         self.g = g
-        self.GRAVITY = g * self.env.OVERRIDDEN_QUAD_MASS # The gravitational force (g*M) acting on each drone.
+        # self.GRAVITY = g * self.env.OVERRIDDEN_QUAD_MASS # The gravitational force (g*M) acting on each drone.
```
Federico-PizarroBejarano, line 64: may as well delete these comments

safe_control_gym/controllers/pid/pid.py
```diff
@@ -259,10 +261,10 @@ def reset_before_run(self, obs=None, info=None, env=None):
         self.last_rpy = np.zeros(3)
         self.integral_rpy_e = np.zeros(3)
 
-        if env is None:
-            self.GRAVITY = self.g * self.env.OVERRIDDEN_QUAD_MASS
-        else:
-            self.GRAVITY = self.g * env.OVERRIDDEN_QUAD_MASS
+        # if env is None:
```
Federico-PizarroBejarano, line 264: Unsure if we should reset the model when the env is reset. I would say yes, since if you added randomization it should trigger whenever it is reset

safe_control_gym/envs/benchmark_env.py
```diff
@@ -298,7 +298,7 @@ def _randomize_values_by_info(self,
         return randomized_values
 
     @abstractmethod
-    def _setup_symbolic(self):
+    def _setup_symbolic(self, prior_prop={}, **kwargs):
```
Federico-PizarroBejarano, line 301: does this return the mode? and the docstring should include args and returns

safe_control_gym/envs/gym_control/cartpole.py
```diff
@@ -392,9 +377,11 @@ def close(self):
             p.disconnect(physicsClientId=self.PYB_CLIENT)
         self.PYB_CLIENT = -1
 
-    def _setup_symbolic(self):
+    def _setup_symbolic(self, prior_prop={}, **kwargs):
```
Federico-PizarroBejarano, line 380: Could copy/paste the complete docstring from benchmarkenv here and in quadrotor

safe_control_gym/envs/gym_control/cartpole.py
```diff
@@ -422,8 +409,18 @@ def _setup_symbolic(self):
         # Define dynamics and cost dictionaries.
         dynamics = {'dyn_eqn': X_dot, 'obs_eqn': Y, 'vars': {'X': X, 'U': U}}
         cost = {'cost_func': cost_func, 'vars': {'X': X, 'U': U, 'Xr': Xr, 'Ur': Ur, 'Q': Q, 'R': R}}
+        # Additional params to cache
+        params = {
```
Federico-PizarroBejarano, line 413: why do we need this?

Federico-PizarroBejarano: ah its passed to SymbolicModel, mb

Federico-PizarroBejarano (changes requested):
Some minor changes and questions, overall looks good and looking forward to testing it with MPSC

safe_control_gym/envs/gym_pybullet_drones/quadrotor.py
```diff
@@ -883,9 +904,16 @@ def _get_reset_info(self):
         '''
         info = {}
         info['symbolic_model'] = self.symbolic
+        # TODO: reverse or delete commented out code once decided how to fix prior
+        # info['physical_parameters'] = {
+        #     'quadrotor_mass': self.MASS,
+        #     'quadrotor_iyy_inertia': self.J[1, 1]
+        # }
         info['physical_parameters'] = {
```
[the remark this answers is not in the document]
Justin-Yuan: I'm not sure why this is logged in the info_dict, don't see any current controllers using it as well. maybe it's legacy code and we can safely remove it.

experiments/pid/config_overrides/quadrotor_stabilization.yaml
Justin-Yuan: oh yes, sorry. this is quick hack to pass the test. i will make a separate ctrl config and modify it in the pid_experiment.sh

safe_control_gym/controllers/base_controller.py
Justin-Yuan: i think adam mentions if multiple controllers want to use the same prior, then it's better to have a yaml for it and use it in `--overrides` so we don't need to copy and paste the prior configs to each ctrl config.   the `mpc.yaml` is an example, i intend to say if there's custom prior for a controller generally we can do `--overrides base.yaml ctrl.yaml prior.yaml` to specify the configs for env, controller and prior

safe_control_gym/controllers/pid/pid.py
Justin-Yuan: ok will do

safe_control_gym/controllers/pid/pid.py
Justin-Yuan: the current randomization only happens once when the ctrl is initialized. i think you are referring to randomizing the prior per epsiode? I'm not sure if this makes sense. since if I imagine a ctrl will learn based on a prior defined/randomized at initialization but fixed throughout training, then if we need to estimate the average performance, we launch multiple training of the controller and each time a training uses one randomized prior. Is there another use case you are thinking of?

safe_control_gym/envs/benchmark_env.py
Justin-Yuan: it doesn't return any, but assign/re-assign the prior model to env.symbolic (so anytime env and ctrl will still have the same prior and env will get the correct quadratic cost from _get_reward()). yse i will add the docstring

safe_control_gym/envs/gym_control/cartpole.py
Justin-Yuan: adam mentions he wanted to move `X_EQ, U_EQ` inside the prior, and since PID need access to the prior quad mass, i just group them inside the prior

safe_control_gym/envs/gym_pybullet_drones/quadrotor.py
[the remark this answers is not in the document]
Justin-Yuan: the prior values can be fetched from the prior model. so here the pysical ones are the actual params used in simulation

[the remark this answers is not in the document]
Federico-PizarroBejarano: I guess there should be a 'safe' way to retrieve the true inertial parameters other than directly accessing the env (which Jacopo doesnt seem to like) so I guess this makes sense. can get it from info.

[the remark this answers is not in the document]
Justin-Yuan: actually i think we can keep this, say if we ever wanna test how domain randomization + prior works, then since each episode will have a differet set of params, it's useful to be able to access them in the info dict

safe_control_gym/controllers/base_controller.py
Federico-PizarroBejarano: Ah ok makes sense!

safe_control_gym/envs/gym_pybullet_drones/quadrotor.py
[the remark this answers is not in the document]
Justin-Yuan: yah. if u agree then i will remove the commented/legacy part then

safe_control_gym/controllers/pid/pid.py
Federico-PizarroBejarano: Ah i guess I was thinking in a more classical sense. For me, when ctrl.reset() is called, its a completely different experiment, so everything should be randomized again, but I see why it would be preferable to have a more explicit way to re-randomizing the prior. How do the inertial parameters get randomized? only once when the environment is initialized?

safe_control_gym/envs/gym_control/cartpole.py
Federico-PizarroBejarano: Makes sense, ty

safe_control_gym/controllers/base_controller.py
```diff
@@ -127,3 +127,67 @@ def load(self,
...
+        prior_prop_rand_info = prior_info.get("prior_prop_rand_info", {})
+        if randomize_prior_prop and prior_prop_rand_info:
+            # check keys, this is due to the current implementation of BenchmarkEnv._randomize_values_by_info()
+            for k in prior_prop_rand_info:
+                assert k in prior_prop, "A prior param to randomize does not have a base value in prior_prop."
+            prior_prop = env._randomize_values_by_info(prior_prop, prior_prop_rand_info)
+
+        # Note we only reset the symbolic model when prior_prop is nonempty
+        if prior_prop:
+            prior_model = env._setup_symbolic(prior_prop=prior_prop)
```
Federico-PizarroBejarano, line 188: if env.setup_symbolic returns nothing, this line is wrong

safe_control_gym/controllers/pid/pid.py
Justin-Yuan: oh i see, you mean doing prior randomization in ctrl.reset().  i thought you mean in ctrl.reset_before_run() since that's called with env.reset().

Federico-PizarroBejarano: Right yeah env.reset() makes more sense than env.reset_before_run(). Would there be a problem in reseting the prior there?

Justin-Yuan: the inertial prop can either be randomized externally and then instantiate env via `inertial_prop`, or it can instantiate env via `inertial_prop_rand_dict` to randomize internally (in the env, so each env.reset() will re-randomize). I use the former case to test robustness of ctrl against different inertial props and the latter case to do domain randomization training.

safe_control_gym/controllers/base_controller.py
Justin-Yuan: oh, sry i forget to remove the `prior_model = ` part, it should just call `env._setup_symbolic`. the comming line `prior_model = env.symbolic` will do the job

safe_control_gym/envs/gym_pybullet_drones/quadrotor.py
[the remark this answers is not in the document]
Federico-PizarroBejarano: yeah i think that makes sense, i hate lingering comments lmao feels cleaner to just delete legacy stuff so yeah leggo

safe_control_gym/controllers/pid/pid.py
Federico-PizarroBejarano: Makes sense. if inertial prop is randomized on every reset, then so should the prior rt

Justin-Yuan: i believe u mean `Right yeah ctrl.reset() makes more sense than ctrl.reset_before_run().`?  

right, i think randomizing prior in `ctrl.reset()` also makes more sense, but i concern that some ctrl might need to use the prior to initialize things for training in `__init__`.  i guess we might somehow need to call `get_prior()` twice, one in `ctrl.__init__` and one in `ctrl.reset()` ?

Federico-PizarroBejarano: Yeah I think thats fine

Justin-Yuan: also on the env randomization of inertial prop, i think randomizing for each episode is only useful for domain randomization. i dont think other controllers need this feature.

Federico-PizarroBejarano: I am planning on using it when I run 100 tests, each with 10% randomization. For those tests I want it to be random on every singular experiment

Justin-Yuan: are you randomizing the `prior_prop` used for ctrl or the `inertail_prop` used for env simulation?

Federico-PizarroBejarano: just randomizing prior_prop. Keep inertial_prop constant and vary how wrong my model is in different ways.

Justin-Yuan: for every singular experiment i think you need prior randomization in `ctrl.reset()`, and if it's for `inertial_prop`, i think u need to randomize the inertial_prop then assign it to `inertial_prop` to the env in each experiment. Do I understand it correctly?

Federico-PizarroBejarano: Wait Im not sure. My understanding would be I would set `inertial_prop_randomization` to None or False or whatever, and set `prior_prop_randomization` to True or a dict or whatever and then whenever I call Experiment.reset() it will reset the env (not changing the inertial prop) and reset the controller (which will randomize the prior). Does that make sense?

Justin-Yuan: yes exactly. so now we are missing the prior randomization in `ctrl.reset()`

Federico-PizarroBejarano: yeah

Justin-Yuan: ok. thought about that a bit. since in that case, we can just call another `get_prior()` in `ctrl.reset()`. i guess if your controller needs it u can just do that in your `ctrl.reset()` function?  the current `get_prior()` will return a optionally randomized prior (from its config), so it will be up to each controller where to call it?

Federico-PizarroBejarano: Yeah that makes sense, so it may as well go into pid.reset rn rt

Justin-Yuan: so it will be up to the user's intension, if you need the prior to construct say a GP or NN in `ctrl.__init__`, you instantiate the prior there; but if you plan to reuse an existing ctrl multiple times (by calling `ctrl.reset()`), you can call `get_prior()` inside `ctrl.reset()`. Would this work?

Justin-Yuan: oh i see, you wanted it that way in PID, ok

Federico-PizarroBejarano: ty!

Justin-Yuan: ok. done. please check here [link](https://github.com/utiasDSL/safe-control-gym/pull/77/commits/888571da9ca0f0fe012306e8db8719b8c2e4cc29) and see if it does what you need?

Federico-PizarroBejarano: Yeah that looks great, thank you!

Justin-Yuan: ok. i will proceed to merging then?

Federico-PizarroBejarano (approved):
LGTM

safe_control_gym/controllers/pid/pid.py
Federico-PizarroBejarano: Yeah I think so, thank you!

PR #79, opened 2022-08-28
Tall and low gates

JacopoPan:


spencerteetaert:
Works on linux machine. Trajectory's first waypoint is at high gate height, and takeoff is only to 0.75m, causing a jump in setpoint in the transition from takeoff to cmdfullstate. Will wait until test on competition machine before merge

Issue #62, opened 2022-08-10
Understanding BoundedConstraint class


nicholasprayogo:
I'm not sure if this is a bug or i'm using the functionality wrong, so please let me know accordingly.

Here is how I'm trying to use BoundedConstraint, specified in my `task_config` `yaml` file. 

```
constraints:
  - constraint_form: bounded_constraint
    lower_bounds: [0, 0, 0] # should match state dim
    upper_bounds: [2.6, 0, 0] 
    constrained_variable: state
    active_dims: 0  # only position
```

When i use [BoundedConstraint](https://github.com/utiasDSL/safe-control-gym/blob/c031b74ea3b05d4c91ca13d4f2fd9cb410d23a70/safe_control_gym/envs/constraints.py#L270) for constraining state (I have 3 states but want to constraint only 1), I realize I needed to supply `lower_bounds` & `upper_bounds` with shapes that are equal to the number of states (e.g. 3 if have 3 states), due to `self.dim` being defined as [such](https://github.com/utiasDSL/safe-control-gym/blob/c031b74ea3b05d4c91ca13d4f2fd9cb410d23a70/safe_control_gym/envs/constraints.py#L56): `self.dim = env.state_dim` which is used in `self.constraint_filter = np.eye(self.dim)[active_dims]` [here](https://github.com/utiasDSL/safe-control-gym/blob/c031b74ea3b05d4c91ca13d4f2fd9cb410d23a70/safe_control_gym/envs/constraints.py#L74), where it is supposed to only extract `active_dims` from these bounds for the target state to be constrained.

But when i do so, where `lower_bounds` matches shape of `env.state_dim`, the assertion [here](https://github.com/utiasDSL/safe-control-gym/blob/c031b74ea3b05d4c91ca13d4f2fd9cb410d23a70/safe_control_gym/envs/constraints.py#L252) `assert A.shape[1] == self.dim, '[ERROR] A has the wrong dimension!'` fails.

This seem to fail because in `Constraint` , inside this [code chunk](https://github.com/utiasDSL/safe-control-gym/blob/c031b74ea3b05d4c91ca13d4f2fd9cb410d23a70/safe_control_gym/envs/constraints.py#L55), after `constraint_filter` is defined, `self.dim` is overwritten by `len(active_dims)`, so it would have shape of `active_dims` which when I use is 1, while `A` already has shape `(6, 3)` due to `self.dim` being set to `env.state_dim` which was 3 earlier. 

```
[...]

PR #88, opened 2022-09-01
Constraint Evaluation Rounding Errors

adamhall:
This branch and PR are for modifying the constraint class to properly handle violations that are a result of numerical issues (as per #85). It was observed that violations can occur on the order of 1e-10 due to the numerical precision of different types the bounds, matrices, and values can take on.

### DIFF
--- a/safe_control_gym/envs/benchmark_env.py
+++ b/safe_control_gym/envs/benchmark_env.py
@@ -372,7 +372,7 @@ class BenchmarkEnv(gym.Env, ABC):
         '''
         # Add initial constraint info (no action/input yet, so only state-based constraints)
         info['current_step'] = 0
-        if self.constraints is not None:
+        if self.constraints is not None and not(self.constraints.state_constraints == []):
             info['constraint_values'] = self.constraints.get_values(self, only_state=True)
         return obs, info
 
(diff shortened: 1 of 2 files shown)


adamhall:
The heart of the issue is that the `DefaultConstraint` class takes it's bounds from the environment's `acction_space` and `state_space`, which cast their upper and lower bounds as `float32`, whereas everything else in SCG is `float64`. `acction_space` and `state_space`, use `float32` because the original openAI gym used `float32` for these. Thus, when the controller thinks it's on the boundary (using a `float64`), its actual evaluation against the constraint (a `float32`), results in constraint violations on the order of 1-e10 to 1e-9. These are clearly the result of numerical issues. To fix this we have a couple of options:

1. Upcasting the `DefaultConstraint` class bounds to `float64`. This actually doesn't work and still results in numerical issues because, although the type is not `float64`, the precision still remains at the `float32` level.
2. Downcasting everything to `float32`. This will work, but contradicts everything else in safe control gym which uses `float64`. Additionally, even if we do this casting, fix 3 (below) should be done anyways to give the user control over the precision they desire in the constraint violations.
3. Round the constraint evaluation using `np.round_` to a specified precision using the new argument `decimals` (same name that numpy uses). I think this is the best solution because it gives the user control over the constraint precision, and avoids casting everything to `float32`.

I've implemented the thrid option in this PR.

JacopoPan (commented):
LGTM, I was going to recommend to also address #62 (which you can also close) and I see you already did. In general, I'd use `try`, `raise`, `except` blocks for "normal" code (over `assert`, reserved for debugging) but up to you.

JacopoPan (approved):

JacopoPan:
Shall this be merged (and the branch deleted?)

PR #90, opened 2022-09-06
Updating MPC classes with new priors structure

adamhall:
This PR is for updating the MPC classes with the new prior setup as per PR #77. Also, I found a bug with the setting of U_EQ that wouldn't update with new mass parameters. Still need to update GPMPC.

### DIFF
--- a/safe_control_gym/controllers/lqr/ilqr.py
+++ b/safe_control_gym/controllers/lqr/ilqr.py
@@ -196,7 +196,7 @@ class iLQR(BaseController):
 
         # Initialize backward pass.
         state_k = self.state_stack[-1]
-        input_k = env.model.U_EQ
+        input_k = env.symbolic.U_EQ
 
         if env.TASK == Task.STABILIZATION:
             x_goal = self.x_0
@@ -205,7 +205,7 @@ class iLQR(BaseController):
         loss_k = loss(x=state_k,
                       u=input_k,
                       Xr=x_goal,
-                      Ur=env.model.U_EQ,
+                      Ur=env.symbolic.U_EQ,
                       Q=self.Q,
                       R=self.R)
         s = loss_k['l'].toarray()
@@ -232,7 +232,7 @@ class iLQR(BaseController):
             loss_k = loss(x=state_k,
                           u=input_k,
                           Xr=x_goal,
-                          Ur=env.model.U_EQ,
+                          Ur=env.symbolic.U_EQ,
                           Q=self.Q,
                           R=self.R)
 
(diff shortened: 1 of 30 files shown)


safe_control_gym/controllers/lqr/ilqr.py
```diff
@@ -21,6 +21,7 @@ def __init__(
             # Model args
             q_lqr: list = None,
             r_lqr: list = None,
+            prior_info: dict={},
```
Federico-PizarroBejarano, line 24: If you leave prior_info in kwargs and call super()__init__ it will set it. I changed baseController slightly in my own PR to set a default to prior prop as well. Not sure if this is the best way to do it, but it means it doesnt need to be added explicitly to each controller and is handled completely by BaseController. We can chat

safe_control_gym/controllers/lqr/ilqr.py
```diff
@@ -57,13 +59,14 @@ def __init__(
         self.env = env_func(info_in_reset=True, done_on_out_of_bound=True)
 
         # Controller params.
-        self.model = self.env.symbolic
+        self.prior_info = prior_info
+        self.model = self.get_prior(self.env)
```
Federico-PizarroBejarano, line 63: in a convo w justin I think we decided that self.get_prior should be called in reset() (as well as init if thats easier). Essentially this is b/c if there is prior randomization it should re-randomize every time the controller or safety filter is reset.

safe_control_gym/controllers/lqr/ilqr.py
```diff
@@ -193,7 +196,7 @@ def update_policy(self, env):
 
         # Initialize backward pass.
         state_k = self.state_stack[-1]
-        input_k = env.U_GOAL
+        input_k = env.symbolic.U_EQ
```
Federico-PizarroBejarano, line 199: shouldnt all of these be self.model.U_EQ?

Federico-PizarroBejarano (changes requested):

safe_control_gym/controllers/lqr/ilqr.py
Justin-Yuan: i think here `self.model` is used to get other information to initialize the controller (e.g. `nx`, `nu`, etc). But Federico is right, feel free to call `get_prior()` in either `__init__()` or `reset()` that you find fit for the specific controller.

safe_control_gym/controllers/lqr/ilqr.py
Justin-Yuan: I agree ideally we don't need to add the `prior_info` argument to every controller's __init__ args. I looked at Federico's PR, that works. Also even without any changes the current `get_prior()` will use the correct defaults but may not be explicit to the user unless they check out the function.

safe_control_gym/controllers/lqr/ilqr.py
adamhall: I think so yes. This is similar to issue #93 and if we want to allow controllers to have access to the true linearization point or only the one provided by the prior. I changed them to `self.model.U_EQ` for now.

adamhall:
Merged as all comments addressed.

PR #92, opened 2022-09-06
Adding priors to MPSC and CBF

Federico-PizarroBejarano:
Adding Justin's prior stuff to MPSC and CBF.


Justin-Yuan:
The changes look good. Also a quick notice that you don't need to add the `prior_info` field in all the YAMLs if it's the default values. Since we call `BenchmarkEnv._setup_symbolic()` for every env initialization, it would already have the default prior model, and calling `BaseController.get_prior()` will just fetch that default prior. This might save some repetitive fields in the config files and ensure backward compatibility with our old experiments.  

But anyway not a big issue, if you test the controllers and they work as expected, then please proceed :)

Justin-Yuan (approved):

Federico-PizarroBejarano:
> 

I mostly did this so that its clear for new users how to change the prior, and so that its easier for me for my future experiments since I plan to change the prior.

Justin-Yuan:
sure. that'd be good as well.

JacopoPan (approved):
In principle LGTM, one suggestion I'd have is to add style/lint'ing changes in separate commits (even if in the same PR) w.r.t. the one with functional changes so that it's easier to review them separately

Federico-PizarroBejarano:
> In principle LGTM, one suggestion I'd have is to add style/lint'ing changes in separate commits (even if in the same PR) w.r.t. the one with functional changes so that it's easier to review them separately

Fair enough, I have an auto-linter set up which makes it somewhat hard to do things separately but I will keep this in mind

PR #108, opened 2022-10-13
10.13 Pull Version

ZQ2413262560:
@JacopoPan  Hello Jacopo!
Our team made a submission, but this submission is not the final version. The primary purpose of this submission is to verify whether our idea is correct, that is, whether we correctly use variables and methods. I only submitted the demo at level2, where we used num_episodes: 1200  in training. Meanwhile, in order to facilitate the test, I also submitted the trained model (I don't know if this is allowed), and we will iterate several versions later. If the version submitted this time does not violate the rules, I believe we can timely provide the final version.
Looking forward to your reply, thank you!


JacopoPan:
Hi @ZQ2413262560 

a high value of `num_episodes` is not forbidden but will likely incur a larger penalty in the "learning efficiency" scoring w.r.t. solutions that use fewer episodes. I also appreciate/encourage the submission of the trained model for the sim2real experiments (otherwise I might just not be able to physically run the training IRL by the announcement of the results).

I will try to run your submission between today and tomorrow. I would appreciate if you could add the command line you expect us to run. Otherwise, I will simply try:
```
cd competition/
python3 getting_started.py --overrides level0.yaml
python3 getting_started.py --overrides level1.yaml
python3 getting_started.py --overrides level2.yaml
python3 getting_started.py --overrides level3.yaml
```
and then take `edit_this.py` into the sim2real setup in the lab

ZQ2413262560:
yeah, just running like `python3 getting_started.py --overrides level0.yaml` is ok. 
If there is a problem with the code we submitted in any way, please let me know, thanks~

utiasDSLadmin:
Hi @ZQ2413262560 (this is @JacopoPan)

I can confirm I can run `python3 getting_started.py --overrides level0.yaml` from the machine where we will generate the evaluations. Just a few notes, I initially received error
```
ModuleNotFoundError: No module named 'cv2'
```
I simply run `conda install opencv` but let me know if you want me to use a specific release/distribution.

Over the weekend, I will look further into how you modified `edit_this`, the YAML files, and `getting_started` to make sure it's all ok. Sending a pre-trained model (along side the code used to generate it is not only permitted but encouraged for the sim2real final evaluation).

I would also appreciate if you could comment here how much time each level (`python3 getting_started.py --overrides level0.yaml`, etc.) takes on your machine(s) to best schedule next week evaluations.

JacopoPan:
@ZQ2413262560 

I looked at your `edit_this.py`, `getting_started.py`, and `level0-3.yaml` files and, on first sight, they look ok.
I noted you set some of the `done` conditions to `False` in the YAML, that is ok for training but not the final evaluation.
If I have doubts in running/scoring the code, I will be in touch over the course of next week. Thank you.

PR #110, opened 2022-10-17
Dev experiment class gpmpc

adamhall:
Modifications to experiment class to facilitate GPMPC training. The main change is that I separated execute_evaluation to have a subfunction execute_task that takes in a specific env and ctrl this makes it more flexible and usable for training.

The same as PR #100 but am trying to make the diff nicer.


adamhall:
Changes explanations:
- Added dir safe_control_gym/experiments to house the different experiment classes
- Added safe_control_gym/experiments/epoch_experiment.py for epoch-based experiments
- Renamed Experiment to BaseExperiment and then changed imports in examples/quad_3D.py and tracking.py, experiments/lqr/lqr_experiment.py, experiments/pid/pid_experiment.py
- an example of how the epoch training is used is given in experiments/arxiv/quadrotor_constraint/utils/DELENDA_epoch_experiment_example.py
- All other files have very small changes or none wrt to the actual dev-experiment-class branch head and all previous experiment class functionality should be the same.

Justin-Yuan (approved):

JacopoPan (changes requested):
Make new PR

PR #119, opened 2022-12-14
MSE counter should be incremented by one?

adamhall:
I noticed that the MSE computation is off by one step. In `step`, the state gets updated, but `self.ctrl_step_counter` doesn't get updated until after `_get_info` and _get_reward` are called. Thus, the `state_error` is subtracting the previous time step's reference state from the current state.


Justin-Yuan:
I don't recall how we define the MSE, but I thought at any given step, aren't we supposed to produce an action/input that tracks a reference state from the given current state? so the error should be the updated state after digesting the action with the reference state?  

another supporting observation is the reference trajectory has shape (T,*), but our state sequence has shape (T+1,*) including the initial and last state (`s_0` and `s_T`). so i think when `s_0` tries to track `ref_0`, we should compute MSE based on the resulting `s_1` and `ref_0`?

adamhall:
Yes this can make sense, but the reference we are actually using starts at t=0. So that means the first point in the reference has generally been close to the initial state of all 0s. So it would be more appropriate to maybe start the reference (as @JacopoPan mentioned on slack) to start the reference at `1/self.CTRL_FREQ`? However, I think its nice having the desired initial state in the reference, for debugging, plotting and for general understanding (like when you want to start on or near the reference).

Justin-Yuan:
i think both points (start the 1st ref to t != 0 and starting near the initial ref) make sense, but it might require modifying the ref traj generation instead of adding 1 to `ctrl_step_counter`? since using `ctrl_step_counter+1` in `get_reward()` and `get_info()` will change the convention, meaning `get_obs()` will need to fetch the next reference with `ctrl_step_counter+2` (in which case at the last timestep it will try to fetch ref[T+1,3], out of bound).

adamhall:
kk I'll think on it a bit to see the best way to do that.

Justin-Yuan (changes requested):
My understanding is this offset to `self.EPISODE_LEN_SEC` extends the reference trajectory by 1 so there is a placeholder initial reference state (not tracked but for visualization or plotting)? correct me if I'm wrong. However, I'd suggest to apply this change to the following https://github.com/utiasDSL/safe-control-gym/blob/05b2817c7573fa14a4bbb23417b5e89e072025df/safe_control_gym/envs/gym_pybullet_drones/quadrotor.py#L277  The reason being BenchmarkEnv's `episode_len_sec` is given by the task config/user and is likely to be used by a controller (if needed), so modifying it (especially in this base class) may be  dangerous. But I think changing it when we generate the reference trajectory (like in line 277 above) makes more sense since that will be our agreed **new convention of the reference**, and it does not affect anything other than the `traj_tracknig` task for the `quadrotor` environment.

Justin-Yuan (changes requested):
For the changes on `ctrl_step_counter`, I assume the reference trajectory is of length (T+1, *) instead of (T, *) now? I think the increments in `_get_reward()` and `_get_info()` make sense, but in `_get_observation()`, I think there is still a bug. 

Specifically, when we first call `env.reset()`, `_get_observation()` should use `ref[1]` which correspond to `ctrl_step_counter+1` (by this time `ctrl_step_counter=0`); and later on in `env.step()`, `_get_observation()` indeed uses `ctrl_step_counter+2`. This is due to we initialize the counter as 0 at reset, but we don't increment it until the end of  the first call to `env.step()`. So I'm afraid the call of `_get_observation()` will be different in `env.reset()` and `env.step()`. 

The proposed change (using `obs = self.extend_obs(obs, self.ctrl_step_counter+2)`) should work for calling `env.step()`, but we might need to have a `if-else` block to make it `obs = self.extend_obs(obs, self.ctrl_step_counter+1)` for calling `env.reset()`. To distinguish which is being called, I think we might need another variable like `self.at_reset` which is set to True in `env.before_reset()`, then it is set to False in `env.after_reset()`. Then inside `_get_observation()`, this variable can be used to know which counter increment to use.

JacopoPan (changes requested):
- don't edit `EPISODE_LENGTH`, make the reference longer by 1
- make sure that, if the index to track the reference has to be shifted, it happens equally in `quadrotor` and `cartpole`

safe_control_gym/envs/gym_control/cartpole.py
```diff
@@ -584,8 +584,15 @@ def _get_observation(self):
...
         if self.obs_wrap_angle:
             obs[2] = normalize_angle(obs[2])
+
         # Concatenate goal info (goal state(s)) for RL
-        obs = self.extend_obs(obs, self.ctrl_step_counter+1)
+        # Plus two because ctrl_step_counter has not incremented yet, and we want to return the obs (which would be
+        # ctrl_step_counter + 1 as the action has already been applited), and the next state (+ 2) for the RL to see
+        # the next state.
+        if self.at_reset:
+            obs = self.extend_obs(obs, self.ctrl_step_counter+1)
```
adamhall, line 593: Make this ```obs = self.extend_obs(obs, 1)```

safe_control_gym/envs/gym_control/cartpole.py
```diff
@@ -205,7 +205,7 @@ def __init__(self,
             self.X_GOAL = np.hstack([self.TASK_INFO['stabilization_goal'][0], 0., 0., 0.])  # x = {x, x_dot, theta, theta_dot}.
         elif self.TASK == Task.TRAJ_TRACKING:
             POS_REF, VEL_REF, _ = self._generate_trajectory(traj_type=self.TASK_INFO['trajectory_type'],
-                                              traj_length=self.EPISODE_LEN_SEC,
+                                              traj_length=self.EPISODE_LEN_SEC + self.CTRL_TIMESTEP, # Added to increment the trajectory by 1 (PR #119)
```
adamhall, line 208: Instead increase generate trajectory by ```times = np.arange(0, traj_length + sample_time, sample_time)```. Double check that this is one longer than before.

JacopoPan: Yes, I think the point is that for an episode with 1 second, 2Hz control frequency, we want "1 reset, 2 control steps" and the reference should be of size 3.

So I would fully encapsulate the mismatch by 1 between the reference and the control steps in the reference generation method.

safe_control_gym/envs/gym_control/cartpole.py
JacopoPan: @Justin-Yuan are we correctly interpreting that the `extend_obs` method takes `self.ctrl_step_counter+2` as second argument except on reset when it takes always `1` (i.e. indexes the second element in the reference)?

safe_control_gym/envs/gym_pybullet_drones/quadrotor.py
```diff
@@ -274,7 +274,7 @@ def __init__(self,
                 ])  # x = {x, x_dot, y, y_dot, z, z_dot, phi, theta, psi, p_body, q_body, r_body}.
         elif self.TASK == Task.TRAJ_TRACKING:
             POS_REF, VEL_REF, _ = self._generate_trajectory(traj_type=self.TASK_INFO['trajectory_type'],
-                                              traj_length=self.EPISODE_LEN_SEC,
+                                              traj_length=self.EPISODE_LEN_SEC + self.CTRL_TIMESTEP, # Added to extend trajectory by 1 (PR #119)
```
JacopoPan, line 277: This wouldn't happen is the ref is extended in the method it is generated

safe_control_gym/envs/gym_pybullet_drones/quadrotor.py
```diff
@@ -798,7 +798,13 @@ def _get_observation(self):
             obs = self.disturbances['observation'].apply(obs, self)
 
         # Concatenate goal info (references state(s)) for RL.
-        obs = self.extend_obs(obs, self.ctrl_step_counter+1)
+        # Plus two because ctrl_step_counter has not incremented yet, and we want to return the obs (which would be
+        # ctrl_step_counter + 1 as the action has already been applited), and the next state (+ 2) for the RL to see
+        # the next state.
+        if self.at_reset:
+            obs = self.extend_obs(obs, self.ctrl_step_counter+1)
```
JacopoPan, line 805: Again, I think the second argument would always evaluate to just 1

JacopoPan (changes requested):

safe_control_gym/envs/gym_control/cartpole.py
adamhall: Added.

safe_control_gym/envs/gym_pybullet_drones/quadrotor.py
adamhall: Fixed.

safe_control_gym/envs/gym_pybullet_drones/quadrotor.py
adamhall: Fixed.

safe_control_gym/experiments/base_experiment.py
```diff
@@ -422,8 +422,13 @@ def get_episode_data(self, key, postprocess_func=lambda x: x):
             episode_data = [postprocess_func(ep_val) for ep_val in self.data[key]]
         elif key in self.data['info'][0][-1]:
             # if the data field is contained in step info dict
-            episode_data = [postprocess_func([info.get(key, 0.) for info in ep_info])
-                            for ep_info in self.data['info']]
+            episode_data = []
+            for ep_info in self.data['info']:
+                ep_info_data = []
+                for info in ep_info:
+                    if key in info:
```
adamhall, line 429: Add verbose flag and warn user if key is not there.

JacopoPan (approved):

adamhall:
@Justin-Yuan do the changes look good to you?

safe_control_gym/envs/benchmark_env.py
```diff
@@ -555,7 +558,7 @@ def _generate_trajectory(self,
         else:
             raise ValueError('Trajectory plane should be in form of ab, where a and b can be {x, y, z}.')
         # Generate time stamps.
-        times = np.arange(0, traj_length, sample_time)
+        times = np.arange(0, traj_length+sample_time, sample_time) #sample time added to make reference one step longer than
```
Federico-PizarroBejarano, line 561: One step longer than what

safe_control_gym/envs/gym_control/cartpole.py
```diff
@@ -584,8 +584,15 @@ def _get_observation(self):
         # Wrap angle to constrain state space, useful in swing-up task.
         if self.obs_wrap_angle:
             obs[2] = normalize_angle(obs[2])
+
         # Concatenate goal info (goal state(s)) for RL
-        obs = self.extend_obs(obs, self.ctrl_step_counter+1)
+        # Plus two because ctrl_step_counter has not incremented yet, and we want to return the obs (which would be
+        # ctrl_step_counter + 1 as the action has already been applited), and the next state (+ 2) for the RL to see
```
Federico-PizarroBejarano, line 590: applied spelled incorrectly

safe_control_gym/envs/gym_pybullet_drones/quadrotor.py
```diff
@@ -798,7 +798,13 @@ def _get_observation(self):
             obs = self.disturbances['observation'].apply(obs, self)
 
         # Concatenate goal info (references state(s)) for RL.
-        obs = self.extend_obs(obs, self.ctrl_step_counter+1)
+        # Plus two because ctrl_step_counter has not incremented yet, and we want to return the obs (which would be
+        # ctrl_step_counter + 1 as the action has already been applited), and the next state (+ 2) for the RL to see
```
Federico-PizarroBejarano, line 802: same typo

safe_control_gym/experiments/base_experiment.py
```diff
@@ -422,8 +429,15 @@ def get_episode_data(self, key, postprocess_func=lambda x: x):
...
+                        ep_info_data.append(info.get(key))
+                    elif self.verbose:
+                        print(f"[Warn] MetricExtractor.get_episode_data: key {key} not in info dict.")
```
Federico-PizarroBejarano, line 439: single quotes if possible lol jk just my personal style choice

Federico-PizarroBejarano (approved):
some small typos but otherwise sounds good

PR #122, opened 2023-01-31
dev-experiment-class minor updates

Federico-PizarroBejarano:
A collection of minor changes I am using to the base software across my branches. I am not sure we should merge all of these, but I believe we should discuss most of them. Let me know your thoughts. 

 - I thought by creating the baseExperiment class we were removing main.py. Unsure if that is still the plan. 
 - Adding seed to reset (and thus experiment classes) is Adam's idea. I can see the benefit. 
 - The structural changes to EpochExperiment are Adam's work for the benchmarking paper
 - The aesthetic changes to EpochExperiment are my changes, including changing the name. 
   - Personally I think EpochExperiment is too messy and too specific to the benchmarking paper to be added to the main branch of the repo. Many of the changes in the dev-experiment-class branch are specific to the paper and should be left in a separate branch, while most of the structural changes should be merged to main imo. If anything, I think the main branch of the repo should be kept as simple and clear as possible people using it (including us) can easily understand whats happening and build upon the code for their own purposes, rather than trying to create a code-base that can be used out-of-the-box.


JacopoPan (changes requested):
Summary of conversation on February 3rd with @Federico-PizarroBejarano @adamhall @Justin-Yuan 
- put back `main.py` and rename it to show it supports RL experiments
- double check that the set and re-set of a seed produces the same experiments 
- move epoch stuff to MPC only experiments (alternatively delete all epch stuff and delenda stuff)

Federico-PizarroBejarano:
1. Fix dev-experiment-class with all minor changes
 - Federico's PR
 - Justin's PR
2. Branch out of dev-experiment-class (called 'benchmark-paper')
 - Add all paper experiments/code here
3. Clean up dev-experiment-class to not have benchmark stuff (ie. epoch experiment)
4. Merge dev-experiment-class

Federico-PizarroBejarano:
> Summary of conversation on February 3rd with @Federico-PizarroBejarano @adamhall @Justin-Yuan
> 
>     * put back `main.py` and rename it to show it supports RL experiments
> 
>     * double check that the set and re-set of a seed produces the same experiments
> 
>     * move epoch stuff to MPC only experiments (alternatively delete all epch stuff and delenda stuff)

1. I put main.py back with a different name, we can work on the name together. 
2. I checked the reseeding and I am pretty sure it works, although depending on how the experiment is set up it could circumvent the reseeding.
3. Is the current plan to merge everything into dev-experiment-class, branch out for the benchmark, then cut everything from dev-experiment-class and set it to main?

Also, I fixed the pytest dependency as Amrit said.

PR #126, opened 2023-04-05
Moving compute_metrics to MetricExtractor

Federico-PizarroBejarano:
Very minor change that I think is useful. I found that the metric extractor class is useful, but right now its only really accessible through the experiment. However, for my experiments I save the trajectory data and want to be able to compute the metrics on the fly. This minor change should make that easier, without changing the interface for the experiments. We could also make the MetricExtractor class completely static, but I tried and its pretty ugly. Not a necessary change but I think its useful.


JacopoPan (approved):
No objection from my side.

PR #131, opened 2023-05-16
Improving Examples and README

Federico-PizarroBejarano:
- Updated README with more examples
- Make training scripts more readable
- Improved PID example as random initialization of state was too random and trajectory tracking was failing.


JacopoPan (approved):
LGTM

PR #161, opened 2024-09-12
MPC warm-starting and acados implementation

MingxuanChe:
This pull request has the recent updates for the MPC controller. 

It focuses on accelerating the online MPC runtime, including warm-starting and an acados implementation. 

1. MPC: 
- update optimizer setup with different solver options.
- add initial guess computation for MPC warm-starting

2. LinearMPC:
- update for solver setup compatibility

3. MPC_ACADOS:
- add initial implementation 
- add example config
- update README about acados-related additional requirement

4. utils and mpc_utils
- add timing decorator
- add MPC open-loop solution plotting

5. Miscellaneous
- colored print-outs for MPC controllers and timing deco

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -150,6 +150,4 @@ dmypy.json
 
 
 # acados
-ilqr_contraction
-arxiv
 acados_ocp_nlp.json
--- a/.gitignore
+++ b/.gitignore
@@ -151,3 +151,4 @@ dmypy.json
 
 # acados
 acados_ocp_nlp.json
+*c_generated_code
--- a/.gitignore
+++ b/.gitignore
@@ -147,3 +147,9 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+
+# acados
+ilqr_contraction
+arxiv
+acados_ocp_nlp.json
--- a/examples/mpc/config_overrides/quadrotor_3D/mpc_acados_quadrotor_3D_stabilization.yaml
+++ b/examples/mpc/config_overrides/quadrotor_3D/mpc_acados_quadrotor_3D_stabilization.yaml
@@ -1,0 +1,28 @@
+algo: mpc
+algo_config:
+  horizon: 20
+  r_mpc:
+    - 0.1
+    - 0.1
+    - 0.1
+    - 0.1
+  q_mpc:
+    - 5.0
+    - 0.1
+    - 5.0
+    - 0.1
+    - 5.0
+    - 0.1
+    - 0.1
+    - 0.1
+    - 0.1
+    - 0.1
+    - 0.1
+    - 0.1
+  # Prior info
+  prior_info:
+    prior_prop: null
+    randomize_prior_prop: False
+    prior_prop_rand_info: null
+  warmstart: True
+  use_RTI: False
--- a/examples/mpc/config_overrides/quadrotor_3D/mpc_acados_quadrotor_3D_stabilization.yaml
+++ b/examples/mpc/config_overrides/quadrotor_3D/mpc_acados_quadrotor_3D_stabilization.yaml
@@ -1,4 +1,4 @@
-algo: mpc
+algo: mpc_acados
 algo_config:
   horizon: 20
   r_mpc:
--- a/safe_control_gym/controllers/mpc/linear_mpc.py
+++ b/safe_control_gym/controllers/mpc/linear_mpc.py
@@ -11,11 +11,13 @@ from sys import platform
 
 import casadi as cs
 import numpy as np
+from termcolor import colored
 
 from safe_control_gym.controllers.lqr.lqr_utils import discretize_linear_system
 from safe_control_gym.controllers.mpc.mpc import MPC
 from safe_control_gym.controllers.mpc.mpc_utils import compute_discrete_lqr_gain_from_cont_linear_system
 from safe_control_gym.envs.benchmark_env import Task
+from safe_control_gym.utils.utils import timing
 
 
 class LinearMPC(MPC):
@@ -36,6 +38,7 @@ class LinearMPC(MPC):
             # shared/base args
             output_dir='results/temp',
             additional_constraints=None,
+            use_lqr_gain_and_terminal_cost: bool = False,
             **kwargs):
         '''Creates task and controller.
 
@@ -51,6 +54,7 @@ class LinearMPC(MPC):
             solver (str): Specify which solver you wish to use (qrqp, qpoases, ipopt, sqpmethod)
             output_dir (str): output directory to write logs and results.
             additional_constraints (list): list of constraints.
+            use_lqr_gain_and_terminal_cost (bool): Use LQR ancillary gain and terminal cost in the MPC.
         '''
         # Store all params/args.
         for k, v in locals().items():
@@ -69,6 +73,8 @@ class LinearMPC(MPC):
             # prior_info=prior_info,
             output_dir=output_dir,
             additional_constraints=additional_constraints,
+            use_lqr_gain_and_terminal_cost=use_lqr_gain_and_terminal_cost,
+            compute_initial_guess_method='lqr',  # use lqr initial guess by default
             **kwargs
         )
 
@@ -105,36 +111,22 @@ class LinearMPC(MPC):
                                                 [x_dot_lin],
                                                 ['x0', 'p'],
                                                 ['xf'])
+        self.lqr_gain, _, _, self.P = compute_discrete_lqr_gain_from_cont_linear_system(dfdx,
+                                                                                        dfdu,
+                                                                                        self.Q,
+                                                                                        self.R,
+                                                                                        self.dt)
         self.dfdx = dfdx
         self.dfdu = dfdu
 
-    def compute_initial_guess(self, init_state, goal_states, x_lin, u_lin):
-        '''Use LQR to get an initial guess of the '''
-        dfdxdfdu = self.model.df_func(x=x_lin, u=u_lin)
-        dfdx = dfdxdfdu['dfdx'].toarray()
-        dfdu = dfdxdfdu['dfdu'].toarray()
-        lqr_gain, _, _ = compute_discrete_lqr_gain_from_cont_linear_system(dfdx, dfdu, self.Q, self.R, self.dt)
-
-        x_guess = np.zeros((self.model.nx, self.T + 1))
-        u_guess = np.zeros((self.model.nu, self.T))
-        x_guess[:, 0] = init_state
-
-        for i in range(self.T):
-            u = lqr_gain @ (x_guess[:, i] - goal_states[:, i]) + u_lin
-            u_guess[:, i] = u
-            x_guess[:, i + 1, None] = self.linear_dynamics_func(x0=x_guess[:, i], p=u)['xf'].toarray()
-
-        return x_guess, u_guess
-
-    def setup_optimizer(self):
+    def setup_optimizer(self, solver='qrqp'):
         '''Sets up convex optimization problem.
-
         Including cost objective, variable bounds and dynamics constraints.
         '''
         nx, nu = self.model.nx, self.model.nu
         T = self.T
         # Define optimizer and variables.
-        if self.solver in ['qrqp', 'qpoases']:
+        if solver in ['qrqp', 'qpoases']:
             opti = cs.Opti('conic')
         else:
             opti = cs.Opti()
@@ -165,7 +157,7 @@ class LinearMPC(MPC):
                           u=np.zeros((nu, 1)) + self.U_EQ[:, None],
                           Xr=x_ref[:, -1],
                           Ur=np.zeros((nu, 1)),
-                          Q=self.Q,
+                          Q=self.Q if not self.use_lqr_gain_and_terminal_cost else self.P,
                           R=self.R)['l']
         for i in range(self.T):
             # Dynamics constraints.
@@ -206,7 +198,7 @@ class LinearMPC(MPC):
         opts = {'expand': True}
         if platform == 'linux':
             opts.update({'print_time': 1, 'print_header': 0})
-            opti.solver(self.solver, opts)
+            opti.solver(solver, opts)
         elif platform == 'darwin':
             opts.update({'ipopt.max_iter': 100})
             opti.solver('ipopt', opts)
@@ -222,6 +214,7 @@ class LinearMPC(MPC):
             'cost': cost
         }
 
+    @timing
     def select_action(self,
                       obs,
                       info=None
@@ -261,19 +254,20 @@ class LinearMPC(MPC):
             self.u_prev = u_val
             self.results_dict['horizon_states'].append(deepcopy(self.x_prev) + self.X_EQ[:, None])
             self.results_dict['horizon_inputs'].append(deepcopy(self.u_prev) + self.U_EQ[:, None])
-        except RuntimeError as e:
-            print(e)
+        except RuntimeError:
+            print(colored('Infeasible MPC Problem', 'red'))
             return_status = opti.return_status()
+            print(colored(f'Optimization failed with status: {return_status}', 'red'))
             if return_status == 'unknown':
-                self.terminate_loop = True
-                u_val = self.u_prev
-                if u_val is None:
-                    print('[WARN]: MPC Infeasible first step.')
-                    u_val = np.zeros((1, self.model.nu))
-            elif return_status == 'Maximum_Iterations_Exceeded':
-                self.terminate_loop = True
-                u_val = opti.debug.value(u_var)
-            elif return_status == 'Search_Direction_Becomes_Too_Small':
+                # self.terminate_loop = True
+                if self.u_prev is None:
+                    print(colored('[WARN]: MPC Infeasible first step.', 'yellow'))
+                    u_val = np.zeros((self.model.nu, self.T))
+                    x_val = np.zeros((self.model.nx, self.T + 1))
+                else:
+                    u_val = self.u_prev
+                    x_val = self.x_prev
+            elif return_status in ['Infeasible_Problem_Detected', 'Infeasible_Problem']:
                 self.terminate_loop = True
                 u_val = opti.debug.value(u_var)
 
@@ -283,5 +277,7 @@ class LinearMPC(MPC):
         else:
             action = np.array([u_val[0]])
         action += self.U_EQ
+        if self.use_lqr_gain_and_terminal_cost:
+            action += self.lqr_gain @ (obs - x_val[:, 0])
         self.prev_action = action
         return action
--- a/safe_control_gym/controllers/mpc/linear_mpc.py
+++ b/safe_control_gym/controllers/mpc/linear_mpc.py
@@ -39,6 +39,7 @@ class LinearMPC(MPC):
             output_dir='results/temp',
             additional_constraints=None,
             use_lqr_gain_and_terminal_cost: bool = False,
+            compute_initial_guess_method=None,
             **kwargs):
         '''Creates task and controller.
 
@@ -74,7 +75,7 @@ class LinearMPC(MPC):
             output_dir=output_dir,
             additional_constraints=additional_constraints,
             use_lqr_gain_and_terminal_cost=use_lqr_gain_and_terminal_cost,
-            compute_initial_guess_method='lqr',  # use lqr initial guess by default
+            compute_initial_guess_method=compute_initial_guess_method,
             **kwargs
         )
 
@@ -254,7 +255,8 @@ class LinearMPC(MPC):
             self.u_prev = u_val
             self.results_dict['horizon_states'].append(deepcopy(self.x_prev) + self.X_EQ[:, None])
             self.results_dict['horizon_inputs'].append(deepcopy(self.u_prev) + self.U_EQ[:, None])
-        except RuntimeError:
+        except RuntimeError as e:
+            print(e)
             print(colored('Infeasible MPC Problem', 'red'))
             return_status = opti.return_status()
             print(colored(f'Optimization failed with status: {return_status}', 'red'))
--- a/safe_control_gym/controllers/mpc/linear_mpc.py
+++ b/safe_control_gym/controllers/mpc/linear_mpc.py
@@ -150,14 +150,14 @@ class LinearMPC(MPC):
             cost += cost_func(x=x_var[:, i] + self.X_EQ[:, None],
                               u=u_var[:, i] + self.U_EQ[:, None],
                               Xr=x_ref[:, i],
-                              Ur=np.zeros((nu, 1)),
+                              Ur=self.U_EQ,
                               Q=self.Q,
                               R=self.R)['l']
         # Terminal cost.
         cost += cost_func(x=x_var[:, -1] + self.X_EQ[:, None],
                           u=np.zeros((nu, 1)) + self.U_EQ[:, None],
                           Xr=x_ref[:, -1],
-                          Ur=np.zeros((nu, 1)),
+                          Ur=self.U_EQ,
                           Q=self.Q if not self.use_lqr_gain_and_terminal_cost else self.P,
                           R=self.R)['l']
         for i in range(self.T):
--- a/safe_control_gym/controllers/mpc/linear_mpc.py
+++ b/safe_control_gym/controllers/mpc/linear_mpc.py
@@ -31,6 +31,7 @@ class LinearMPC(MPC):
             r_mpc=[1],
             warmstart=True,
             soft_constraints=False,
+            soft_penalty: float = 10000,
             terminate_run_on_done=True,
             constraint_tol: float = 1e-8,
             solver: str = 'sqpmethod',
@@ -56,6 +57,7 @@ class LinearMPC(MPC):
             output_dir (str): output directory to write logs and results.
             additional_constraints (list): list of constraints.
             use_lqr_gain_and_terminal_cost (bool): Use LQR ancillary gain and terminal cost in the MPC.
+            compute_initial_guess_method (str): Method to compute the initial guess for the MPC. Options: None, 'ipopt', 'lqr'.
         '''
         # Store all params/args.
         for k, v in locals().items():
@@ -69,6 +71,7 @@ class LinearMPC(MPC):
             r_mpc=r_mpc,
             warmstart=warmstart,
             soft_constraints=soft_constraints,
+            soft_penalty=soft_penalty,
             terminate_run_on_done=terminate_run_on_done,
             constraint_tol=constraint_tol,
             # prior_info=prior_info,
@@ -80,9 +83,6 @@ class LinearMPC(MPC):
         )
 
         # TODO: setup environment equilibrium
-        # self.X_EQ = np.atleast_2d(self.env.X_GOAL)[0,:].T
-        # self.U_EQ = np.atleast_2d(self.env.U_GOAL)[0,:]
-
         self.X_EQ = np.atleast_2d(self.model.X_EQ)[0, :].T
         self.U_EQ = np.atleast_2d(self.model.U_EQ)[0, :].T
         assert solver in ['qpoases', 'qrqp', 'sqpmethod', 'ipopt'], '[Error]. MPC Solver not supported.'
@@ -166,7 +166,7 @@ class LinearMPC(MPC):
             opti.subject_to(x_var[:, i + 1] == next_state)
 
             # State and input constraints
-            soft_con_coeff = 10
+            soft_con_coeff = self.soft_penalty
             for sc_i, state_constraint in enumerate(self.state_constraints_sym):
                 if self.soft_constraints:
                     opti.subject_to(state_constraint(x_var[:, i] + self.X_EQ.T) <= state_slack[sc_i])
@@ -270,7 +270,7 @@ class LinearMPC(MPC):
                     u_val = self.u_prev
                     x_val = self.x_prev
             elif return_status in ['Infeasible_Problem_Detected', 'Infeasible_Problem']:
-                self.terminate_loop = True
+                # self.terminate_loop = True
                 u_val = opti.debug.value(u_var)
 
         # take first one from solved action sequence
--- a/safe_control_gym/controllers/mpc/mpc.py
+++ b/safe_control_gym/controllers/mpc/mpc.py
@@ -253,14 +253,14 @@ class MPC(BaseController):
             cost += cost_func(x=x_var[:, i],
                               u=u_var[:, i],
                               Xr=x_ref[:, i],
-                              Ur=np.zeros((nu, 1)),
+                              Ur=self.U_EQ,
                               Q=self.Q,
                               R=self.R)['l']
         # Terminal cost.
         cost += cost_func(x=x_var[:, -1],
                           u=np.zeros((nu, 1)),
                           Xr=x_ref[:, -1],
-                          Ur=np.zeros((nu, 1)),
+                          Ur=self.U_EQ,
                           Q=self.Q if not self.use_lqr_gain_and_terminal_cost else self.P,
                           R=self.R)['l']
         # Constraints
--- a/safe_control_gym/controllers/mpc/mpc.py
+++ b/safe_control_gym/controllers/mpc/mpc.py
@@ -50,6 +50,7 @@ class MPC(BaseController):
             r_mpc (list): diagonals of input/action cost weight.
             warmstart (bool): if to initialize from previous iteration.
             soft_constraints (bool): Formulate the constraints as soft constraints.
+            soft_penalty (float): Penalty added in the cost function for soft constraints.
             terminate_run_on_done (bool): Terminate the run when the environment returns done or not.
             constraint_tol (float): Tolerance to add the the constraint as sometimes solvers are not exact.
             output_dir (str): output directory to write logs and results.
--- a/safe_control_gym/controllers/mpc/mpc.py
+++ b/safe_control_gym/controllers/mpc/mpc.py
@@ -1,16 +1,18 @@
 '''Model Predictive Control.'''
-
 from copy import deepcopy
 
 import casadi as cs
 import numpy as np
+from termcolor import colored
 
 from safe_control_gym.controllers.base_controller import BaseController
+from safe_control_gym.controllers.lqr.lqr_utils import discretize_linear_system
 from safe_control_gym.controllers.mpc.mpc_utils import (compute_discrete_lqr_gain_from_cont_linear_system,
                                                         compute_state_rmse, get_cost_weight_matrix,
                                                         reset_constraints, rk_discrete)
 from safe_control_gym.envs.benchmark_env import Task
 from safe_control_gym.envs.constraints import GENERAL_CONSTRAINTS, create_constraint_list
+from safe_control_gym.utils.utils import timing
 
 
 class MPC(BaseController):
@@ -24,6 +26,7 @@ class MPC(BaseController):
             r_mpc: list = [1],
             warmstart: bool = True,
             soft_constraints: bool = False,
+            soft_penalty: float = 10000,
             terminate_run_on_done: bool = True,
             constraint_tol: float = 1e-6,
             # runner args
@@ -32,6 +35,10 @@ class MPC(BaseController):
             additional_constraints: list = None,
             use_gpu: bool = False,
             seed: int = 0,
+            compute_initial_guess_method: str = 'ipopt',
+            use_lqr_gain_and_terminal_cost: bool = False,
+            init_solver: str = 'ipopt',
+            solver: str = 'ipopt',
             **kwargs
     ):
         '''Creates task and controller.
@@ -49,14 +56,16 @@ class MPC(BaseController):
             additional_constraints (list): List of additional constraints
             use_gpu (bool): False (use cpu) True (use cuda).
             seed (int): random seed.
+            compute_initial_guess_method (str): Method to compute the initial guess. Options: None, 'ipopt', 'lqr'.
+            use_lqr_gain_and_terminal_cost (bool): Use the LQR ancillary gain and terminal cost in the MPC.
+            init_solver (str): Solver to use for initial guess computation.
+            solver (str): Solver to use for MPC optimization.
         '''
         super().__init__(env_func=env_func, output_dir=output_dir, use_gpu=use_gpu, seed=seed, **kwargs)
         for k, v in locals().items():
             if k != 'self' and k != 'kwargs' and '__' not in k:
                 self.__dict__.update({k: v})
 
-        # if prior_info is None:
-        #    self.prior_info = {}
         # Task.
         self.env = env_func()
         if additional_constraints is not None:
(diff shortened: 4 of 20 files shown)


safe_control_gym/controllers/mpc/mpc.py
```diff
@@ -291,14 +348,39 @@ def select_action(self,
...
+                print(f'Optimization failed with status: {return_status}')
+                if return_status == 'unknown':
+                    # self.terminate_loop = True
+                    u_val = self.u_prev
+                    x_val = self.x_prev
+                    if u_val is None:
+                        print('[WARN]: MPC Infeasible first step.')
+                        u_val = u_guess
+                        x_val = x_guess
+                elif return_status == 'Maximum_Iterations_Exceeded':
```
Federico-PizarroBejarano, line 369: these two if statements lead to the same result

safe_control_gym/controllers/mpc/mpc.py
```diff
@@ -291,14 +348,39 @@ def select_action(self,
...
+                    self.terminate_loop = True
+                    u_val = opti.debug.value(u_var)
+                    x_val = opti.debug.value(x_var)
+                elif return_status == 'Search_Direction_Becomes_Too_Small':
+                    self.terminate_loop = True
```
Federico-PizarroBejarano, line 374: I don't quite understand how this works. It seems to affect only the `run` function, but the run function was something we were trying to completely phase out in preference for the experiment class. In fact, I think the run function should be deleted if possible

safe_control_gym/controllers/mpc/mpc.py
```diff
@@ -407,14 +489,14 @@ def run(self,
             self.results_dict['action'].append(action)
             self.results_dict['state'].append(env.state)
             self.results_dict['state_mse'].append(info['mse'])
-            # self.results_dict['state_error'].append(env.state - env.X_GOAL[i,:])
-
+            if self.solver == 'ipopt':
```
Federico-PizarroBejarano, line 492: why only do this when using ipopt?

safe_control_gym/controllers/mpc/mpc.py
```diff
@@ -407,14 +489,14 @@ def run(self,
...
+                self.results_dict['state_error'].append(env.state - env.X_GOAL[i, :])
             common_metric += info['mse']
```
Federico-PizarroBejarano, line 494: common_metric?? I know this wasnt added in this PR but this seems like a super confusing variable. Could we change or delete this to be more clear

safe_control_gym/envs/gym_control/cartpole.py
```diff
@@ -442,8 +442,13 @@ def _set_observation_space(self):
         self.theta_threshold_radians = 90 * math.pi / 180
```
Federico-PizarroBejarano, line 442: Can we reorganize the thresholds to be in the correct order?

safe_control_gym/envs/gym_pybullet_drones/quadrotor.py
```diff
@@ -639,43 +639,49 @@ def _set_observation_space(self):
         self.phi_threshold_radians = 85 * math.pi / 180
         self.theta_threshold_radians = 85 * math.pi / 180
         self.psi_threshold_radians = 180 * math.pi / 180  # Do not bound yaw.
+        self.x_dot_threshold = 15
```
Federico-PizarroBejarano, line 642: same here, can we reorganize the thresholds to be in the same order as they appear in the obs? seems more organized.

safe_control_gym/controllers/mpc/mpc_acados.py
```diff
@@ -0,0 +1,412 @@
...
+            **kwargs
+        )
+
+        self.x_guess = None
+        self.u_guess = None
+        # acados settings
+        self.use_RTI = use_RTI
+
+    @timing
+    def reset(self):
```
Federico-PizarroBejarano, line 92: Some of this is repeated from the MPC class. I would call super to not duplicate code.

safe_control_gym/controllers/mpc/mpc_acados.py
```diff
@@ -0,0 +1,412 @@
...
+        self.setup_acados_model()
+        # Acados optimizer.
+        self.setup_acados_optimizer()
+        self.acados_ocp_solver = AcadosOcpSolver(self.ocp)
+        # Previously solved states & inputs, useful for warm start.
+        self.x_prev = None
+        self.u_prev = None
+        self.setup_results_dict()
+
+    def setup_acados_model(self) -> AcadosModel:
```
Federico-PizarroBejarano, line 119: Add docstrings if possible

safe_control_gym/controllers/mpc/mpc_acados.py
```diff
@@ -0,0 +1,412 @@
...
+        '''
+        # store meta information # NOTE: unit is missing
+        acados_model.x_labels = self.env.STATE_LABELS
+        acados_model.u_labels = self.env.ACTION_LABELS
+        acados_model.t_label = 'time'
+
+        self.acados_model = acados_model
+
+    @timing
+    def compute_initial_guess(self, init_state, goal_states):
```
Federico-PizarroBejarano, line 149: Is this the same as the function in MPC? why is it being overridden?

Federico-PizarroBejarano (changes requested):
Thanks for the PR, in general looks good. I have not worked much with the MPC classes so I am new, but I have a few comments:

1. There already seems to be a ton of duplication in the MPC examples, like the stabilization and tracking yamls being identical. Your acados ones add even more duplication. there has to be a way to make it way simpler
2. Although i don't dislike the way you've done the MPC_ACADOS subclass, in general I dont like the idea of one of the controllers being just a different solver of another one of the controllers. Not only is it not really a separate controller, but it adds a lot of duplicated code. In my own code I made acados work by adding separate setup and solving functions in the same class: https://github.com/Federico-PizarroBejarano/safe-control-gym/blob/99142ae57663c23aec204931e38aadc06883bdf6/safe_control_gym/safety_filters/mpsc/mpsc.py#L121. If you think this would be uglier than your current approach, that is fine, just something to consider. 
3. How does this PR affect GP_MPC? I think the renaming of the initial guess function directly impacts it, although i have never used it and I assume you know better. Just wanted to ensure there are no unintended consequences

adamhall:
> 1. There already seems to be a ton of duplication in the MPC examples, like the stabilization and tracking yamls being identical. Your acados ones add even more duplication. there has to be a way to make it way simpler

I wonder if for the examples the duplication is okay, simply because if someone wants to use a very specific function of the gym, they can immediately see an example for that function, as opposed to having to understand why all three examples are the same?

> 2. Although i don't dislike the way you've done the MPC_ACADOS subclass, in general I dont like the idea of one of the controllers being just a different solver of another one of the controllers. Not only is it not really a separate controller, but it adds a lot of duplicated code. In my own code I made acados work by adding separate setup and solving functions in the same class: https://github.com/Federico-PizarroBejarano/safe-control-gym/blob/99142ae57663c23aec204931e38aadc06883bdf6/safe_control_gym/safety_filters/mpsc/mpsc.py#L121. If you think this would be uglier than your current approach, that is fine, just something to consider.

This is a good point Fed, but I think I am mostly okay with it being a separate controller. ACADOS is more than just a different solver, its an entirely different parser built on top of CasADi specifically for MPC giving access to more MPC-specific solvers. On the other hand, the structure you have is cleaner overall IMO.

> 3. How does this PR affect GP_MPC? I think the renaming of the initial guess function directly impacts it, although i have never used it and I assume you know better. Just wanted to ensure there are no unintended consequences

There is another PR for GPMPC with ACADOS nearly ready, we wanted to get this one in first.

Federico-PizarroBejarano:
@adamhall  understood. I am happy to leave those three issues as they are in the PR

safe_control_gym/controllers/mpc/mpc.py
```diff
@@ -115,8 +121,9 @@ def close(self):
         self.env.close()
 
     def reset(self):
-        '''Prepares for training or evaluation.'''
```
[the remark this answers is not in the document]
MingxuanChe: This is moved back to its original place in the latest commit.

safe_control_gym/controllers/mpc/mpc.py
```diff
@@ -137,38 +144,86 @@ def reset(self):
...
+        delta_x = cs.MX.sym('delta_x', self.model.nx, 1)
+        delta_u = cs.MX.sym('delta_u', self.model.nu, 1)
+        Ad, Bd = discretize_linear_system(dfdx, dfdu, self.dt, exact=True)
+        x_dot_lin = Ad @ delta_x + Bd @ delta_u
+        self.linear_dynamics_func = cs.Function('linear_discrete_dynamics',
+                                                [delta_x, delta_u],
+                                                [x_dot_lin],
+                                                ['x0', 'p'],
+                                                ['xf'])
+        self.dfdx = dfdx
```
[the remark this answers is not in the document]
MingxuanChe: This linearized prior dynamics will be used together with linearized GP dynamics to compute the LQR anciliary gain in GP-MPC. Sorry this is not clear in this PR.

safe_control_gym/controllers/mpc/mpc.py
```diff
@@ -137,38 +144,86 @@ def reset(self):
...
+        x_init = opti_dict['x_init']  # initial state
+        x_ref = opti_dict['x_ref']  # reference state/trajectory
+
+        # Assign the initial state.
+        opti.set_value(x_init, init_state)  # initial state should have dim (nx,)
+        # Assign reference trajectory within horizon.
+        goal_states = self.get_references()
+        opti.set_value(x_ref, goal_states)
+        if self.mode == 'tracking':
+            self.traj_step += 1
```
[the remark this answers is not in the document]
MingxuanChe: This is fixed in the latest commit: compute intial guess won't increase the traj_step and only select_action will.

safe_control_gym/controllers/mpc/mpc.py
```diff
@@ -137,38 +144,86 @@ def reset(self):
...
             u = lqr_gain @ (x_guess[:, i] - goal_states[:, i]) + u_lin
             u_guess[:, i] = u
             x_guess[:, i + 1, None] = self.dynamics_func(x0=x_guess[:, i], p=u)['xf'].toarray()
 
         return x_guess, u_guess
 
-    def setup_optimizer(self):
+    @timing
+    def compute_initial_guess(self, init_state, goal_states):
+        '''Use IPOPT to get an initial guess of the '''
```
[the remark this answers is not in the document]
MingxuanChe: The ideal is to use IPOPT to compute an initial guess to the MPC problem and then switch to an SQP solver (here I take QRSQP as an example). The reasons are as follow:
1. when tested in a cartpole stabilization task, swithing to SQP solver after IPOPT initial guess is faster than always using IPOPT.
2. I observe more failure when only using SQP solver and using IPOPT increases the success rate. But the success rate is still not satisfactory on 2D/3D quadrotor; that's also why I need acados.

safe_control_gym/controllers/mpc/mpc.py
```diff
@@ -277,11 +332,13 @@ def select_action(self,
         opti.set_value(x_ref, goal_states)
         if self.mode == 'tracking':
             self.traj_step += 1
-        # if self.warmstart and self.x_prev is None and self.u_prev is None:
-        #    x_guess, u_guess = self.compute_initial_guess(obs, goal_states, self.X_EQ, self.U_EQ)
-        #    opti.set_initial(x_var, x_guess)
-        #    opti.set_initial(u_var, u_guess) # Initial guess for optimization problem.
-        # elif self.warmstart and self.x_prev is not None and self.u_prev is not None:
+
+        if self.compute_ipopt_initial_guess and self.x_prev is None and self.u_prev is None:
```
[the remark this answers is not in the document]
MingxuanChe: Currently no. In the latest commit I merged the lqr and ipopt initial guess to one function.

safe_control_gym/controllers/mpc/mpc.py
```diff
@@ -291,14 +348,39 @@ def select_action(self,
             opti.set_initial(x_var, x_guess)
             opti.set_initial(u_var, u_guess)
         # Solve the optimization problem.
-        sol = opti.solve()
-        x_val, u_val = sol.value(x_var), sol.value(u_var)
+        try:
+            sol = opti.solve()
+            x_val, u_val = sol.value(x_var), sol.value(u_var)
+        except RuntimeError:
```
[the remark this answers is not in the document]
MingxuanChe: There are two reasons:
1. For hardware safety: when MPC is running on the hardware, throwing an error due to infeasiblity will let the hardware repeat the last avaible command, which usually leads to dangerous behaviour. 
2. For HPO: @middleyuan mentioned previously, throwing an error during HPO can lead to worse HP sampling due to the setup. 
We can talk about it more later.

safe_control_gym/controllers/mpc/mpc.py
```diff
@@ -291,14 +348,39 @@ def select_action(self,
...
+            print(colored('Infeasible MPC Problem', 'red'))
+            if self.solver == 'ipopt':
+                x_val, u_val = opti.debug.value(x_var), opti.debug.value(u_var)
+            elif self.solver == 'qrsqp':
+                return_status = opti.return_status()
+                print(f'Optimization failed with status: {return_status}')
```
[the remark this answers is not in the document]
MingxuanChe: This is fixed in the latest commit.

safe_control_gym/controllers/mpc/mpc.py
```diff
@@ -277,11 +332,13 @@ def select_action(self,
...
-        # if self.warmstart and self.x_prev is None and self.u_prev is None:
-        #    x_guess, u_guess = self.compute_initial_guess(obs, goal_states, self.X_EQ, self.U_EQ)
-        #    opti.set_initial(x_var, x_guess)
-        #    opti.set_initial(u_var, u_guess) # Initial guess for optimization problem.
-        # elif self.warmstart and self.x_prev is not None and self.u_prev is not None:
+
+        if self.compute_ipopt_initial_guess and self.x_prev is None and self.u_prev is None:
+            #    x_guess, u_guess = self.compute_lqr_initial_guess(obs, goal_states, self.X_EQ, self.U_EQ)
+            print(colored(f'computing initial guess with {self.init_solver}', 'green'))
+            x_guess, u_guess = self.compute_initial_guess(obs, goal_states)
```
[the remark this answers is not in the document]
MingxuanChe: Thank you for pointing this out. This is fixed in the latest commit.

safe_control_gym/controllers/mpc/mpc.py
MingxuanChe: This is fixed in the latest commit.

safe_control_gym/controllers/mpc/mpc.py
MingxuanChe: This is fixed in the latest commit.

safe_control_gym/envs/gym_pybullet_drones/quadrotor.py
MingxuanChe: This reordering is done in the lates commit.

safe_control_gym/controllers/mpc/mpc_acados.py
MingxuanChe: Thank you for this suggestion. I was not aware I can do super previously and now super is added.

safe_control_gym/controllers/mpc/mpc_acados.py
MingxuanChe: This is updated with using super to reduce some duplicated parts

safe_control_gym/controllers/mpc/linear_mpc.py
```diff
@@ -222,6 +223,7 @@ def setup_optimizer(self):
             'cost': cost
         }
 
+    @timing
```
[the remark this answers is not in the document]
MingxuanChe: This is done by calling the function in a wrapper and doing timing in the wrapper [[link](https://github.com/MingxuanChe/safe-control-gym/blob/6104a5d3f0f22e9c0e38c2f9dfdada92b24f5712/safe_control_gym/utils/utils.py#L201)].

MingxuanChe:
Thank you both for the comments! I tried to address the most of the comments and there are still several terms I might need some further discussion.

The remaining itesm for this PR:
1. whether we remove .run methods
2. whether we delete common_metric
3. handeling when MPC solver fails (throw an error or not?)

safe_control_gym/controllers/mpc/mpc.py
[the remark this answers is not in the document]
Federico-PizarroBejarano: This could always be moved to the GP PR then but its fine

safe_control_gym/controllers/mpc/mpc.py
```diff
@@ -137,38 +144,86 @@ def reset(self):
...
+        self.dfdx = dfdx
+        self.dfdu = dfdu
+
         self.dynamics_func = rk_discrete(self.model.fc_func,
                                          self.model.nx,
                                          self.model.nu,
                                          self.dt)
 
-    def compute_initial_guess(self, init_state, goal_states, x_lin, u_lin):
+    def compute_lqr_initial_guess(self, init_state, goal_states, x_lin, u_lin):
```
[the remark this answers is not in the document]
Federico-PizarroBejarano: I am assuming this is going to be used in the GP stuff somehow?

safe_control_gym/controllers/mpc/mpc.py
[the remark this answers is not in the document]
Federico-PizarroBejarano: Seems cool, how often do you compute the initial guess?

safe_control_gym/controllers/mpc/mpc.py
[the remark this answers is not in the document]
Federico-PizarroBejarano: Ah I saw your next comment

safe_control_gym/controllers/mpc/mpc.py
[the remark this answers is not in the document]
Federico-PizarroBejarano: Are the debug values safe? What I do in the safety filter stuff is fall back on an LQR controller when things fail. I'm ok with whatever you think is best for HPO and real life tho

Federico-PizarroBejarano:
Hmmm unfortunately I guess its hard to add tests since it would require installing ACADOS on GitHub :(. Annoying, can't think of a good work around

safe_control_gym/controllers/mpc/linear_mpc.py
```diff
@@ -69,6 +73,8 @@ def __init__(
             # prior_info=prior_info,
             output_dir=output_dir,
             additional_constraints=additional_constraints,
+            use_lqr_gain_and_terminal_cost=use_lqr_gain_and_terminal_cost,
+            compute_initial_guess_method='lqr',  # use lqr initial guess by default
```
Federico-PizarroBejarano, line 77: Why would this not be set by the parameters inherited from the MPC class? It should not always be LQR rt, it should be a parameter that can be passed in like all the others? If there is a default, it needs to be set on the MPC init func rt?

safe_control_gym/controllers/mpc/linear_mpc.py
```diff
@@ -165,7 +157,7 @@ def setup_optimizer(self):
                           u=np.zeros((nu, 1)) + self.U_EQ[:, None],
                           Xr=x_ref[:, -1],
                           Ur=np.zeros((nu, 1)),
-                          Q=self.Q,
+                          Q=self.Q if not self.use_lqr_gain_and_terminal_cost else self.P,
```
Federico-PizarroBejarano, line 160: would be interested in learning more about this

safe_control_gym/controllers/mpc/linear_mpc.py
```diff
@@ -261,19 +254,20 @@ def select_action(self,
             self.u_prev = u_val
             self.results_dict['horizon_states'].append(deepcopy(self.x_prev) + self.X_EQ[:, None])
             self.results_dict['horizon_inputs'].append(deepcopy(self.u_prev) + self.U_EQ[:, None])
-        except RuntimeError as e:
-            print(e)
+        except RuntimeError:
```
Federico-PizarroBejarano, line 257: why not print out the error `e`?

safe_control_gym/controllers/mpc/mpc_utils.py
```diff
@@ -90,3 +92,59 @@ def reset_constraints(constraints):
     if len(constraints_list.input_state_constraints) > 0:
         raise NotImplementedError('[Error] Cannot handle combined state input constraints yet.')
     return constraints_list, state_constraints_sym, input_constraints_sym
+
+
+def set_acados_constraint_bound(constraint,
+                                bound_type,
+                                bound_value=None,
+                                ):
+    '''Set the acados constraint bound.
```
Federico-PizarroBejarano, line 101: ideally add full args and returns for new functions. Its kind of a pain but will improve documentation, especially later if we decide to use a documentation engine like sphinx

safe_control_gym/controllers/mpc/mpc_utils.py
```diff
@@ -90,3 +92,59 @@ def reset_constraints(constraints):
...
+        if bound_type == 'lb':
+            bound_value = -1e8
+        elif bound_type == 'ub':
+            bound_value = 1e-6
+
+    return bound_value * np.ones(constraint.shape)
+
+
+def plot_open_loop_sol(ctrl):
+    ''' Plot the open loop predction of the MPC controller.
```
Federico-PizarroBejarano, line 119: prediction is misspelt

Federico-PizarroBejarano (changes requested):
A few more minor comments

safe_control_gym/controllers/mpc/mpc_acados.py
```diff
@@ -0,0 +1,398 @@
...
+                else:
+                    init_u = self.u_guess[:, idx]
+                self.acados_ocp_solver.set(idx, 'u', init_u)
+
+        # set reference for the control horizon
+        goal_states = self.get_references()
+        if self.mode == 'tracking':
+            self.traj_step += 1
+
+        y_ref = np.concatenate((goal_states[:, :-1], np.zeros((nu, self.T))))
```
Federico-PizarroBejarano, line 325: should this be `np.zeros((nu, self.T))` or `self.U_EQ`?

MingxuanChe: Thank you for pointing this out. I agree `self.U_EQ` could make more sense and I will fix this.

safe_control_gym/controllers/mpc/mpc.py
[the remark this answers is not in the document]
MingxuanChe: I usually only use it for the first step.

safe_control_gym/controllers/mpc/linear_mpc.py
MingxuanChe: Thanks for pointing this out. This will be reverted.

safe_control_gym/controllers/mpc/mpc_utils.py
MingxuanChe: Thanks for pointing this out. This will be fixed.

safe_control_gym/controllers/mpc/mpc_utils.py
MingxuanChe: I added full args to the one you pointed out but not sure whether there is still some missing one / old ones without full args. Might will address this more carefully in the future PR.

safe_control_gym/controllers/mpc/linear_mpc.py
MingxuanChe: This is basically using the LQR solution for 1) terminal cost and 2) ancillary gain and to improve the MPC solution quality. But since the terminal constraint set is missing, this wouldn't provide any rigorous theoretical guarantee.

safe_control_gym/controllers/mpc/linear_mpc.py
MingxuanChe: This is a good point. Right now, for the initial guess computation, I have only two options: lqr or using ipopt. The ipopt option will solve a nonlinear MPC step. I think using a nonlinear MPC solution to warm-start a linear MPC optimization makes less sense, but certainly, we can do this. However, what I have here is actually problematic in the sense that it will always start with an lqr initial guess. I will make it a parameter instead of hard-code it like this.

safe_control_gym/controllers/mpc/mpc.py
```diff
@@ -24,6 +26,7 @@ def __init__(
             r_mpc: list = [1],
             warmstart: bool = True,
             soft_constraints: bool = False,
+            soft_penalty: float = 10000,
```
Federico-PizarroBejarano, line 29: Missing this in the docstring

safe_control_gym/controllers/mpc/mpc_acados.py
```diff
@@ -0,0 +1,391 @@
...
+    ):
+        '''Creates task and controller.
+
+        Args:
+            env_func (Callable): function to instantiate task/environment.
+            horizon (int): mpc planning horizon.
+            q_mpc (list): diagonals of state cost weight.
+            r_mpc (list): diagonals of input/action cost weight.
+            warmstart (bool): if to initialize from previous iteration.
+            soft_constraints (bool): Formulate the constraints as soft constraints.
```
Federico-PizarroBejarano, line 57: soft_penalty missing here too

safe_control_gym/controllers/mpc/mpc_acados.py
```diff
@@ -0,0 +1,391 @@
...
+            output_dir=output_dir,
+            additional_constraints=additional_constraints,
+            # compute_initial_guess_method='ipopt',  # use ipopt initial guess by default
+            use_lqr_gain_and_terminal_cost=use_lqr_gain_and_terminal_cost,
+            use_gpu=use_gpu,
+            seed=seed,
+            **kwargs
+        )
+
+        self.x_guess = None
```
Federico-PizarroBejarano, line 89: would these need to be reset? Does it make sense to put these in the `reset` function and then just calling it?

safe_control_gym/controllers/mpc/mpc_acados.py
```diff
@@ -0,0 +1,391 @@
...
+
+        # set up rk4 (acados need symbolic expression of dynamics, not function)
+        k1 = self.model.fc_func(acados_model.x, acados_model.u)
+        k2 = self.model.fc_func(acados_model.x + self.dt / 2 * k1, acados_model.u)
+        k3 = self.model.fc_func(acados_model.x + self.dt / 2 * k2, acados_model.u)
+        k4 = self.model.fc_func(acados_model.x + self.dt * k3, acados_model.u)
+        f_disc = acados_model.x + self.dt / 6 * (k1 + 2 * k2 + 2 * k3 + k4)
+
+        acados_model.disc_dyn_expr = f_disc
+        '''
```
Federico-PizarroBejarano, line 128: I would put this in the function docstring maybe. I found it a bit sudden and out of place here

safe_control_gym/controllers/mpc/mpc_acados.py
```diff
@@ -0,0 +1,391 @@
...
+        # slack costs for nonlinear constraints
+        if self.soft_constraints:
+            # slack variables for all constraints
+            ocp.constraints.Jsh_0 = np.eye(h0_expr.shape[0])
+            ocp.constraints.Jsh = np.eye(h_expr.shape[0])
+            ocp.constraints.Jsh_e = np.eye(he_expr.shape[0])
+            # slack penalty
+            L2_pen = self.soft_penalty
+            L1_pen = self.soft_penalty
+            ocp.cost.Zl_0 = L2_pen * np.ones(h0_expr.shape[0])
```
Federico-PizarroBejarano, line 206: should we soften the initial constraint?

Federico-PizarroBejarano: And are we softening the input constraints? we shouldnt but not sure if we are

safe_control_gym/controllers/mpc/mpc_acados.py
```diff
@@ -0,0 +1,391 @@
...
+        x_init = np.zeros((nx))
+        ocp.constraints.x0 = x_init
+
+        # set up solver options
+        ocp.solver_options.qp_solver = 'PARTIAL_CONDENSING_HPIPM'
+        ocp.solver_options.hessian_approx = 'GAUSS_NEWTON'
+        ocp.solver_options.integrator_type = 'DISCRETE'
+        ocp.solver_options.nlp_solver_type = 'SQP' if not self.use_RTI else 'SQP_RTI'
+        ocp.solver_options.nlp_solver_max_iter = 25 if not self.use_RTI else 1
+        # ocp.solver_options.globalization = 'FUNNEL_L1PEN_LINESEARCH' if not self.use_RTI else 'MERIT_BACKTRACKING'
```
Federico-PizarroBejarano, line 229: as usual, i think commented out code needs to be either deleted or included unless it really serves a purpose

safe_control_gym/controllers/mpc/mpc_acados.py
```diff
@@ -0,0 +1,391 @@
...
+
+    @timing
+    def select_action(self,
+                      obs,
+                      info=None
+                      ):
+        '''Solves nonlinear mpc problem to get next action.
+
+        Args:
+            obs (ndarray): Current state/observation.
```
Federico-PizarroBejarano, line 301: Sometimes we have a period at the end, sometimes not. Unsure what the more common convention is in our code. Should eventually standardize. Not a big deal

Federico-PizarroBejarano (changes requested):
Some minor comments, its looking good I think

safe_control_gym/controllers/mpc/mpc.py
Federico-PizarroBejarano: Still find this wording strange

MingxuanChe: @adamhall Any suggestion on improving this?

.gitignore
```diff
@@ -147,3 +147,9 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+
+# acados
+ilqr_contraction
```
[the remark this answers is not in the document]
MingxuanChe: These are irrelevant actually. I will remove them.

examples/mpc/config_overrides/quadrotor_3D/mpc_acados_quadrotor_3D_stabilization.yaml
```diff
@@ -0,0 +1,28 @@
+algo: mpc
```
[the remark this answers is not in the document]
MingxuanChe: Updated this for the two config files and checked for all the other configs. Should be good now.

safe_control_gym/controllers/mpc/linear_mpc.py
```diff
@@ -51,6 +55,7 @@ def __init__(
             solver (str): Specify which solver you wish to use (qrqp, qpoases, ipopt, sqpmethod)
             output_dir (str): output directory to write logs and results.
             additional_constraints (list): list of constraints.
+            use_lqr_gain_and_terminal_cost (bool): Use LQR ancillary gain and terminal cost in the MPC.
```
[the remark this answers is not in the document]
MingxuanChe: This is added. Thank you for pointing this out.

safe_control_gym/controllers/mpc/linear_mpc.py
```diff
@@ -263,17 +257,19 @@ def select_action(self,
...
-                self.terminate_loop = True
-                u_val = self.u_prev
-                if u_val is None:
-                    print('[WARN]: MPC Infeasible first step.')
-                    u_val = np.zeros((1, self.model.nu))
-            elif return_status == 'Maximum_Iterations_Exceeded':
-                self.terminate_loop = True
-                u_val = opti.debug.value(u_var)
-            elif return_status == 'Search_Direction_Becomes_Too_Small':
+                # self.terminate_loop = True
```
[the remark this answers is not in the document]
MingxuanChe: I agree with your comments. But some of them are more or less undecided about what is the best. I would keep it for now and clean up later.

safe_control_gym/envs/gym_control/cartpole.py
```diff
@@ -438,12 +438,16 @@ def _set_action_space(self):
...
     def _set_observation_space(self):
         '''Sets the observation space of the environment.'''
-        # Angle at which to fail the episode.
-        self.theta_threshold_radians = 90 * math.pi / 180
         # NOTE: different value in PyBullet gym (0.4) and OpenAI gym (2.4).
         self.x_threshold = 2.4
+        self.x_dot_threshold = 10
+        self.theta_threshold_radians = 90 * math.pi / 180  # Angle at which to fail the episode.
+        self.theta_dot_threshold = 10
         # Limit set to 2x: i.e. a failing observation is still within bounds.
```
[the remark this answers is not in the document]
MingxuanChe: That was actually me. I added this purely for acados since acados cannot handle a inf large constraint. https://discourse.acados.org/t/infeasible-qps-when-using-nonlinear-casadi-constraint-expressions/1595/3?u=mxche

safe_control_gym/controllers/mpc/mpc.py
MingxuanChe: This is added now. Thank you!

safe_control_gym/controllers/mpc/mpc_acados.py
MingxuanChe: I agree putting it into reset is better, and I have updated this.

safe_control_gym/controllers/mpc/mpc_acados.py
MingxuanChe: I agree with the initial condition part. For the input constraint part I mainly want to keep the consistency with the current MPC formulation. @adamhall any suggestion on this?

safe_control_gym/controllers/mpc/mpc_acados.py
MingxuanChe: This has been moved to docstring.

safe_control_gym/controllers/mpc/mpc_acados.py
MingxuanChe: I agree and have put this to the reset.

safe_control_gym/controllers/mpc/mpc_acados.py
MingxuanChe: I agree with the initial condition part. For the input constraints I mainly keep the consistency with the current MPC formulation. @adamhall any suggestion on this?

safe_control_gym/controllers/mpc/mpc_acados.py
MingxuanChe: I have this for two reasons 1) for the user to know there are some options to improve the MPC solving; 2) easy for me to try out when needed. So I would like to keep them if this doesn't matter that much.

safe_control_gym/controllers/mpc/mpc_acados.py
MingxuanChe: I checked the acados formulation [(eq (2) and (3))](https://github.com/acados/acados/blob/master/docs/problem_formulation/problem_formulation_ocp_mex.pdf It seems that the initial condition is a separate constraint and would be affected by the softening. The softening will affect the other constraints acting on the initial node and I think this should not affect the optimization much.

safe_control_gym/controllers/mpc/mpc_acados.py
Federico-PizarroBejarano: Do we need to call reset in the init function? Will it throw an error somewhere if x_guess is never set?

safe_control_gym/envs/gym_control/cartpole.py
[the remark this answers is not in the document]
Federico-PizarroBejarano: Sounds good let me know what you decide

.gitignore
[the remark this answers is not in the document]
Federico-PizarroBejarano: what about `c_generated_code` files from acados?

safe_control_gym/controllers/mpc/mpc_acados.py
MingxuanChe: In MPC code, this will only be used for warm-starting, and the none case will be handled [here](https://github.com/MingxuanChe/safe-control-gym/blob/208e844836f3ae487fe2ff85f50dd34a836e0eee/safe_control_gym/controllers/mpc/mpc_acados.py#L316C1-L320C46). I think this should be safe.

.gitignore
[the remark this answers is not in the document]
MingxuanChe: Interestingly, I don't see any `c_generated_code` files but usually there are some. They should be included in the gitignore for sure. I will add them.

safe_control_gym/envs/gym_control/cartpole.py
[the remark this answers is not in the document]
MingxuanChe: I will double them for now, as well as for the quadrotor.

safe_control_gym/controllers/mpc/mpc_acados.py
```diff
@@ -0,0 +1,390 @@
...
+        # placeholder initial state constraint
+        x_init = np.zeros((nx))
+        ocp.constraints.x0 = x_init
+
+        # set up solver options
+        ocp.solver_options.qp_solver = 'PARTIAL_CONDENSING_HPIPM'
+        ocp.solver_options.hessian_approx = 'GAUSS_NEWTON'
+        ocp.solver_options.integrator_type = 'DISCRETE'
+        ocp.solver_options.nlp_solver_type = 'SQP' if not self.use_RTI else 'SQP_RTI'
+        ocp.solver_options.nlp_solver_max_iter = 25 if not self.use_RTI else 1
```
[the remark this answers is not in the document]
adamhall: testing @MingxuanChe

adamhall (changes requested):

safe_control_gym/controllers/mpc/mpc.py
```diff
@@ -49,14 +56,16 @@ def __init__(
             additional_constraints (list): List of additional constraints
             use_gpu (bool): False (use cpu) True (use cuda).
             seed (int): random seed.
+            compute_initial_guess_method (str): Method to compute the initial guess. Options: None, 'ipopt', 'lqr'.
+            use_lqr_gain_and_terminal_cost (bool): Use the LQR ancillary gain and terminal cost in the MPC.
```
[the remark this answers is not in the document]
MingxuanChe: Right now, I don't have a terminal set in the formulation, so with only terminal cost and ancillary gain, it wouldn't really provide any guarantee afaik. We can integrate @Federico-PizarroBejarano ’s robust MPC at some point and I will actually need it.

safe_control_gym/controllers/mpc/mpc_acados.py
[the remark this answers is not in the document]
MingxuanChe: We can have them in yaml for sure. But I might need some more time to thoroughly test the options acados provides. Or we can just put all the available options there and let the user decide.

safe_control_gym/controllers/mpc/mpc_acados.py
```diff
@@ -0,0 +1,390 @@
...
+
+        self.ocp = ocp
+
+    def processing_acados_constraints_expression(self, ocp: AcadosOcp, h0_expr, h_expr, he_expr) -> AcadosOcp:
+        '''Preprocess the constraints to be compatible with acados.
+            Args:
+                ocp (AcadosOcp): acados ocp object
+                h0_expr (casadi expression): initial state constraints
+                h_expr (casadi expression): state and input constraints
+                he_expr (casadi expression): terminal state constraints
```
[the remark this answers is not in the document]
MingxuanChe: Thank you for this suggestion. This makes sense and I will change them.

Federico-PizarroBejarano (approved):
LGTM

PR #164, opened 2024-09-23
Hyperparameter Optimization Module

middleyuan:
This PR is created for two primary purposes:

1. Include the state-of-the-art package Google-Vizier as a hyperparameter optimization solver.
2. Change the database to SQLite for easier usage.


safe_control_gym/hyperparameters/hpo_vizier.py
```diff
@@ -0,0 +1,429 @@
...
+import yaml, csv
+import matplotlib.pyplot as plt
+
+from safe_control_gym.hyperparameters.hpo_sampler import HYPERPARAMS_DICT
+from safe_control_gym.experiments.base_experiment import BaseExperiment
+from safe_control_gym.utils.logging import ExperimentLogger
+from safe_control_gym.utils.registration import make
+from safe_control_gym.utils.utils import mkdirs
+
+class HPO(object):
```
adamhall, line 27: This class is defined in both hpo_optuna and hpo_vizier, naming should be more precise instead of having the same class defined in two places.

adamhall (changes requested):
Looks pretty good! I have a couple thoughts though:

1. It seems like there is a lot of controller-specific code. For example, there is a separate sampler for each controller, and in the HPO code, there are if statements depending on the algorithm being optimized. I'm wondering if there is a way to make this more generic by making an HPO yaml more complex and then having the underlying code make classes from the arguments in these yamls? For example, I think that hpo_sampler.py could almost entirely be defined in a yaml and then having a generic class for sampler that parses the yaml appropriately? It feels like there is a lot of repeated code that could be simplified and the addition of future algos simpler?

2. the HPO class is defined in both hpo_optuna.py and hpo_vizier.py which seem to share a lot of code. I think there should really be a parent HPO class, and then child sublcasses for the different use cases?

3. I'm not totally sure why the files are being removed from the examples. Is it because you are replacing them with better hyper parameters?

safe_control_gym/hyperparameters/hpo_optuna.py
```diff
@@ -0,0 +1,429 @@
...
+        while increase_runs:
+            increase_runs = False
+            if first_iteration:
+                Gs = np.inf
+            for i in range(self.hpo_config.repetitions):
+                # np.random.seed()
+                seed = np.random.randint(0, 10000)
+                # update the agent config with sample candidate hyperparameters
+                # new agent with the new hps
+                for hp in sampled_hyperparams:
```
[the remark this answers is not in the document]
middleyuan: Actually, I am thinking that we should unify the naming of q and r across controllers. Then we could avoid this?

examples/hpo/cartpole/config_overrides/gp_mpc_cartpole_stab_200.yaml
```diff
@@ -49,18 +50,18 @@ algo_config:
...
   rand_data_selection: false
   terminate_train_on_done: True
   terminate_test_on_done: False
+  # parallel: False 
   parallel: True
 
-device: cpu
-restore: null
+device: cuda
+restore: null
```
Federico-PizarroBejarano, line 67: there should be a newline at the end of the file. I think the pre-commit hooks handle that. If not, you can configure vs code to always add one when a file is saved, which is what I do

Federico-PizarroBejarano: Yeah the pre-commit hook should fix this, which means you haven't yet run the hook. This should be done ASAP as it may change a lot

examples/hpo/cartpole/config_overrides/ppo_cartpole_stab_.yaml
```diff
@@ -2,7 +2,7 @@ algo: ppo
 algo_config:
   # model args
   hidden_dim: 64
-  activation: relu
+  activation: "relu"
```
Federico-PizarroBejarano, line 5: quotations are not necessary in yaml

examples/hpo/hpo_experiment.py
```diff
@@ -18,87 +19,75 @@ def hpo(config):
...
-MAIN_FUNCS = {'hpo': hpo, 'train': train}
+MAIN_FUNCS = {'hpo': hpo}
 
 
 if __name__ == '__main__':
     # Make config.
     fac = ConfigFactory()
-    fac.add_argument('--func', type=str, default='train', help='main function to run.')
-    fac.add_argument('--opt_hps', type=str, default='', help='yaml file as a result of HPO.')
+    fac.add_argument('--func', type=str, default='hpo', help='main function to run.')
```
Federico-PizarroBejarano, line 87: what other options are there other than 'hpo'?

Federico-PizarroBejarano (changes requested):
I agree with Adam's comments and believe they should be addressed before the next round of reviews. The changes are too broad to do only one review. In general I think there is too much code, both repeated and non-repeated. I think the gym needs to generally simple and concise for people (other than us) to use it. There is also a lot of commented out code which I don't think should be there in completed code. Also, two stylistic comments: the precommit hooks need to be run to format everything consistently, and the docstrings and comments should follow our standard style, as Adam pointed out.

adamhall:
Looks much cleaner! Nice! I think the samplers could still be made a little more flexable? Let me know if you think this is feasible.

middleyuan:
> Looks pretty good! I have a couple thoughts though:
> 
> 1. It seems like there is a lot of controller-specific code. For example, there is a separate sampler for each controller, and in the HPO code, there are if statements depending on the algorithm being optimized. I'm wondering if there is a way to make this more generic by making an HPO yaml more complex and then having the underlying code make classes from the arguments in these yamls? For example, I think that hpo_sampler.py could almost entirely be defined in a yaml and then having a generic class for sampler that parses the yaml appropriately? It feels like there is a lot of repeated code that could be simplified and the addition of future algos simpler?
> 2. the HPO class is defined in both hpo_optuna.py and hpo_vizier.py which seem to share a lot of code. I think there should really be a parent HPO class, and then child sublcasses for the different use cases?
> 3. I'm not totally sure why the files are being removed from the examples. Is it because you are replacing them with better hyper parameters?

1. I have made the HPO code more generic. The reason I don't define hyperparameter search space in yaml is that I don't want to add burdens on users as it usually requires some knowledge for code and algorithms. 
2.  Yes, corresponding changes are made.
3. Changes include re-factoring to make the folder structure consistent with other examples.

Generally comments: to get HPO module fully tested, I am waiting for another PR (quadrotor interface) to be approved. After that I will run unit-test on new env interface and pre-commits hook.

safe_control_gym/hyperparameters/hpo_optuna.py
```diff
@@ -0,0 +1,429 @@
...
+
+        # do repetition
+        returns, seeds = [], []
+        while increase_runs:
+            increase_runs = False
+            if first_iteration:
+                Gs = np.inf
+            for i in range(self.hpo_config.repetitions):
+                # np.random.seed()
+                seed = np.random.randint(0, 10000)
```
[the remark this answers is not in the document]
middleyuan: Yes, see `hpo_experiment.py`.

safe_control_gym/hyperparameters/hpo_optuna.py
```diff
@@ -0,0 +1,429 @@
...
+from safe_control_gym.utils.registration import make
+from safe_control_gym.utils.utils import mkdirs
+
+
+class HPO(object):
+
+    def __init__(self, algo, task, load_study, output_dir, task_config, hpo_config, algo_config, safety_filter=None, sf_config=None):
+        """ Hyperparameter optimization class
+
+        args:
```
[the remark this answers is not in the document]
middleyuan: Done

safe_control_gym/hyperparameters/hpo_optuna.py
```diff
@@ -0,0 +1,429 @@
...
+            algo: algo name
```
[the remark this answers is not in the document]
middleyuan: Done

safe_control_gym/hyperparameters/hpo_optuna.py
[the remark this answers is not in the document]
middleyuan: Done

safe_control_gym/hyperparameters/hpo_vizier.py
middleyuan: Done

examples/hpo/cartpole/config_overrides/gp_mpc_cartpole_stab_200.yaml
middleyuan: Done

examples/hpo/cartpole/config_overrides/ppo_cartpole_stab_.yaml
middleyuan: Done

examples/hpo/hpo_experiment.py
middleyuan: Deleted it.

examples/hpo/cartpole/config_overrides/gp_mpc_cartpole_stab_200.yaml
Federico-PizarroBejarano: I think the hooks have still not been run, since there are double quotes here and there.

Federico-PizarroBejarano:
Closing for now since the code is likely very outdated. Can open a new one if necessary.

PR #171, opened 2024-10-30
Fix ilqr max trajectory step and updates for hardware compatiblity

MingxuanChe:
This pull request has the recent updates for the iLQR controller.

It mainly 

1. fixed the bug of the hard-coded max trajectory steps, and 
2. added a step counter which will increase every time the `select_action` is called. This will help to run the controller on the hardware.

I am happy to address any questions you have.


MingxuanChe:
I will create a new one once the MPC PR is finalized.

PR #173, opened 2024-11-05
fix ilqr max step and add step counter

MingxuanChe:
This pull request is a cleaned-up version of the previous closed ilqr PR and has the recent updates for the iLQR controller.

The main changes are

1. fixed the bug of the hard-coded max trajectory steps, and
2. added a step counter which will increase every time the `select_action` is called. This will help to run the controller on the hardware: when running the controller on the hardware, the controller will read the open-loop action of ilqr according to this counter, and the counter will be restricted to be no larger than `self.max_step`.
3. updated some docstrings

I am happy to address any questions you have. Also, let me know if the way I restrict the counter increment looks good to you [(here)](https://github.com/MingxuanChe/safe-control-gym/blob/29b18b4df6c24e12063adabddd9bb4bb33abe4ef/safe_control_gym/controllers/lqr/ilqr.py#L311C1-L312C32).

### DIFF
--- a/safe_control_gym/controllers/lqr/ilqr.py
+++ b/safe_control_gym/controllers/lqr/ilqr.py
@@ -1,4 +1,4 @@
-'''Linear Quadratic Regulator (LQR)
+'''iterative Linear Quadratic Regulator (iLQR)
 
 [1] https://studywolf.wordpress.com/2016/02/03/the-iterative-linear-quadratic-regulator-method/
 [2] https://arxiv.org/pdf/1708.09342.pdf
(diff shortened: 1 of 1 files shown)


safe_control_gym/controllers/lqr/ilqr.py
```diff
@@ -39,12 +39,16 @@ def __init__(
...
+            epsilon (float): The convergence tolerance of the cost function.
+
+        Note: This implementation has a Hessian regularization term lambda
+        to make sure the Hessian H is well-conditioned for inversion. See [1] for more details.
         '''
 
         super().__init__(env_func, **kwargs)
 
         # Model parameters
+        self.env = env_func()
```
Federico-PizarroBejarano, line 51: why do you create an self.env here and also on line 62?

safe_control_gym/controllers/lqr/ilqr.py
```diff
@@ -275,29 +285,31 @@ def select_action(self, obs, info=None, training=False):
         Args:
             obs (ndarray): The observation at this timestep.
             info (dict): The info at this timestep.
+            training (bool): Whether the algorithm is training or evaluating.
 
         Returns:
             action (ndarray): The action chosen by the controller.
         '''
 
-        step = self.extract_step(info)
```
Federico-PizarroBejarano, line 283: Not that I am against this change, but why didn't the original implementation work? The `extract_step` function was created specifically to get this info? Are we not passing the info in every time or what?

Federico-PizarroBejarano (changes requested):
Confused about two things

safe_control_gym/controllers/lqr/ilqr.py
Federico-PizarroBejarano: Will open an issue on this

Federico-PizarroBejarano (approved):
LGTM

adamhall (approved):
LGTM

PR #175, opened 2024-11-06
Add acados solver for 3D quadrotor

aghezz1:
Quick comparison against current NMPC controller using CasADi/IPOPT for the task of tracking a figure of 8 with a 3D quadrotor model.

[acados](https://github.com/acados/acados) achieves similar or better performance than IPOPT while dramatically reducing computation time.
Computation time goes down at least a factor 10 for fully converged SQP (tolerance 1e-6), almost a factor 100 for the RTI scheme! 🚀

Table below shows the result analysis with the provided function.
Computation times are in milliseconds.

---
TODOs
- [ ] Automatically parse constraints from safe-control-gym to acados constraints (at the moment hard coded for only one example)

---

| Stats.                |  Casadi/IPOPT | acados SQP | acados SQP RTI |
|-----------------------|---------------|------------|----------------|
| average_length        | 300.000       |   300.000          | 300.000                |
| length                | 300.000       |  300.000           |   300.000            |
| average_return        |-485.747       |  -599.505            |    -544.023            |
| average_rmse          |0.384          |  0.334             |   0.340             |
| rmse                  |0.384          |  0.334             |  0.340              |
| rmse_std              |0.000          |  0.000             |   0.000             |
| worst_case_rmse_at_0.5 |0.384         |  0.334             |   0.340             |
| failure_rate            |0.000        |    0.000           |   0.000             |
| average_constraint_violation |0.000   |   0.000            |    0.000            |
| constraint_violation_std |0.000       |    0.000           |     0.000            |
| constraint_violation  |0.000          |    0.000           |     0.000            |
| **mean_t_wall_ms**        |72.268         |   **14.320**           |    **0.997**            |
| **median_t_wall_ms**      |65.351         |   **5.200**           |   **0.919**          |

### DIFF
--- a/safe_control_gym/controllers/mpc/linear_mpc.py
+++ b/safe_control_gym/controllers/mpc/linear_mpc.py
@@ -87,8 +87,8 @@ class LinearMPC(MPC):
         dfdxdfdu = self.model.df_func(x=self.X_EQ, u=self.U_EQ)
         dfdx = dfdxdfdu['dfdx'].toarray()
         dfdu = dfdxdfdu['dfdu'].toarray()
-        delta_x = cs.MX.sym('delta_x', self.model.nx, 1)
-        delta_u = cs.MX.sym('delta_u', self.model.nu, 1)
+        delta_x = cs.SX.sym('delta_x', self.model.nx, 1)
+        delta_u = cs.SX.sym('delta_u', self.model.nu, 1)
         # x_dot_lin_vec = dfdx @ delta_x + dfdu @ delta_u
         # self.linear_dynamics_func = cs.integrator(
         #    'linear_discrete_dynamics', self.model.integration_algo,
--- a/safe_control_gym/math_and_models/symbolic_systems.py
+++ b/safe_control_gym/math_and_models/symbolic_systems.py
@@ -69,7 +69,7 @@ class SymbolicModel():
         # Discrete time dynamics.
         self.fd_func = cs.integrator('fd', self.integration_algo, {'x': self.x_sym,
                                                                    'p': self.u_sym,
-                                                                   'ode': self.x_dot}, {'tf': self.dt}
+                                                                   'ode': self.x_dot}, 0, self.dt
                                      )
         # Observation model.
         self.g_func = cs.Function('g', [self.x_sym, self.u_sym], [self.y_sym], ['x', 'u'], ['g'])
@@ -88,8 +88,8 @@ class SymbolicModel():
                                    [self.dgdx, self.dgdu], ['x', 'u'],
                                    ['dgdx', 'dgdu'])
         # Evaluation point for linearization.
-        self.x_eval = cs.MX.sym('x_eval', self.nx, 1)
-        self.u_eval = cs.MX.sym('u_eval', self.nu, 1)
+        self.x_eval = cs.SX.sym('x_eval', self.nx, 1)
+        self.u_eval = cs.SX.sym('u_eval', self.nu, 1)
         # Linearized dynamics model.
         self.x_dot_linear = self.x_dot + self.dfdx @ (
             self.x_eval - self.x_sym) + self.dfdu @ (self.u_eval - self.u_sym)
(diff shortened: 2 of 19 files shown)


adamhall:
This is a really cool addition! The timing is pretty interesting as just yesterday we merged a PR to include an MPC using acados (see #161). You can take a look there so see how your implementation lines up, but from a high level they look very similar.

The comparison you have done is really interesting. I'm surprised that ACADOS is not only significantly faster, but also achieves better performance. Naively, I always thought that IPOPT would get better performance, but was usually slower. It would be interesting to compare the robustness of the solvers as well! This is great work!

safe_control_gym/math_and_models/symbolic_systems.py
```diff
@@ -69,7 +69,7 @@ def setup_model(self):
         # Discrete time dynamics.
         self.fd_func = cs.integrator('fd', self.integration_algo, {'x': self.x_sym,
                                                                    'p': self.u_sym,
-                                                                   'ode': self.x_dot}, {'tf': self.dt}
+                                                                   'ode': self.x_dot}, 0, self.dt
```
Federico-PizarroBejarano, line 72: I am interested as why this was changed. It seemed to be working before, although from what I am seeing on the documentation, perhaps the way we were doing it was deprecated? Can you tell me more?

Federico-PizarroBejarano (commented):
Hi @aghezz1, thank you so much for your contribution. It is interesting that we merged in an acados MPC just yesterday, but I think this is a great opportunity to work together to get the best implementation. I would really appreciate any tips on improving our current implementation based on the work you have done here. To that effect, I briefly went through your PR and commented on a few changes that seem interesting but I don't know enough about them. Let me know what you think, and thanks again!!

aghezz1:
Thanks guys for sharing the excitements! and for your nice software package 😄 
Some syscop colleagues and I were at your workshop at CDC23!

The timing is fun, I cloned the repo last Friday and didn't see the PR #161 about acados.
I will have a look in the next days into #161
Overall the two crucial things are:
- automatic parsing of constraints from your class to acados constraints
- you should decide whether extending the class MPC to welcome every solver (I guess this is what you did in #161) or consider MPCAcados as a new controller. Extending the MPC class might become a bit overwhelming but it's only my personal speculation.

safe_control_gym/math_and_models/symbolic_systems.py
aghezz1: Yes, the dict with `tf` was deprecated starting from casadi 3.6.X.
It still works but you get a warning so I updated it.

safe_control_gym/controllers/mpc/linear_mpc.py
```diff
@@ -87,8 +87,8 @@ def set_dynamics_func(self):
         dfdxdfdu = self.model.df_func(x=self.X_EQ, u=self.U_EQ)
         dfdx = dfdxdfdu['dfdx'].toarray()
         dfdu = dfdxdfdu['dfdu'].toarray()
-        delta_x = cs.MX.sym('delta_x', self.model.nx, 1)
-        delta_u = cs.MX.sym('delta_u', self.model.nu, 1)
+        delta_x = cs.SX.sym('delta_x', self.model.nx, 1)
```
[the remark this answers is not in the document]
aghezz1: SX builds "scalar" expression graphs which tends to be less memory efficient but faster for computation.
My rule of thumb is that MX should be used in case you have memory problem or if you need black box function evaluation in your casadi expressions. Btw, when you pass `{"expand": True}` as option to casadi `nlpsol` you are transforming all the MX in your NLP to SX.
Also, SX expression are easier to debug.

safe_control_gym/math_and_models/symbolic_systems.py
Federico-PizarroBejarano: Awesome, thank you!

aghezz1:
One more comment about performance and reliability.
- IPOPT will be more reliable than acados, there are some underlying reasons in the algorithms they use for solving NLPs (happy to talk about them separately).
- I think I misread my results, I guess that your `average_return` metric should be maximized (so the higher the better). If this is the case then IPOPT is a bit better then acados. On the other end, for the `rmse` metric acados is better than IPOPT.

MingxuanChe:
Very cool comparison! I started my implementation in PR #161 based on my understanding of acados and I am happy to learn about your implementation. Let's see how we can integrate them. I will take a closer look in the following days. 

Before that I have two minor comments: 
1. Which acados release you are using? I believe I am using an older version.

aghezz1:
Continuing in #182.

PR #176, opened 2024-11-07
Update the bibtex of the second publication from arXiv to RAL

JacopoPan:
Update the bibtex of the second publication from arXiv to RAL


Federico-PizarroBejarano (approved):
LGTM

PR #181, opened 2024-12-02
Removing set_cost_function_param and setting it via rew_state_weight

Federico-PizarroBejarano:
Previously there was no direct way to set the Q and R matrices that determine the cost when using the Quadratic cost, except a function `set_cost_function_param` called from the controller or experiment. I believe this violates a separation of controller and environment, and the Q and R matrices should be set similarly to the reward for RL_Reward via the environment yaml. To this effect, I have done the following:

1. Deleted `set_cost_function_param` everywhere
2. Defined Q and R in the environment as `np.diag(rew_state_weight)` and `np.diag(rew_act_weight)`
3. Set the appropriate `rew_state_weight` and `rew_act_weight` in the examples such that the environment is has the same cost as the LQR and MPC experiments. However, this is not necessary as in general the cost functions in the controller and environment can be different. 
4. ~~Additionally deleted the `run` function in `mpc.py` as it is unnecessary and has been replaced by the `base_experiment.py` class.~~


aghezz1:
I tested it and works fine, thanks!
I have few remarks:
1. Regarding matrices Q, R: do they refer to the discretized system or to the continuous one? If we consider a simple Euler integrator for the cost, the difference between Q_cont and Q_disc is that Q_disc = dt * Q_cont. In acados, we specify Q and R for the continuous system, then the cost is integrated via Euler (so by multiplying it with the sampling time `dt`). You can have a look to this [document](https://github.com/acados/acados/blob/master/docs/problem_formulation/problem_formulation_ocp_mex.pdf).
2. I agree to delete the function `run(...)` in the file `mpc.py` since I believe is not used. Also, in the files `mpc_acados.py` and `linear_mpc.py` there is no function `run(...)`.
3. I would also remove the training environment from `mpc.py` as there is no training to be done, it is a bit confusing and takes some time to just set an environment which is not used (especially for acados as it needs to compile a new solver). Maybe there is a reason why there is a training environment but I didn't get it.

Unrelated, after our chat #175, I had time to look into your acados MPC controller and saw some things that can be improved. After merging this PR I would open a new one with some edits 😄

Federico-PizarroBejarano:
Awesome!

1. That's very interesting. Up until now we have been specifying the discrete Q and R matrices. I think this aligns well with reinforcement learning since we often consider the per-step reward, but I can see that for classical control it makes sense to choose the continuous Q and R so they don't change if we change the controller frequency. I think for now we can keep it as it is and think more about this and whether it is necessary to make a change. 
2. @MingxuanChe realized last night that the GP-MPC uses the run function in `mpc.py` for training. There is likely a better way to do this, but for now we will leave it as it is and remove it when necessary. 
3. I believe the training environment was added for GP-MPC, but that hasn't been added yet so I agree it is currently superfluous. I am happy to clean this up. Will think a bit more about whether there is any reason to keep this as it currently is. 

Awesome, we would really appreciate your input on the acados stuff. We have been meaning to go more carefully through your previous PR but unfortunately we're pretty busy with some deadlines at the moment. However, if you make a PR into our current acados implementation we will definitely take a look at it, thank you!

examples/lqr/lqr_experiment.py
```diff
@@ -15,11 +15,12 @@
 from safe_control_gym.utils.registration import make
 
 
-def run(gui=True, n_episodes=1, n_steps=None, save_data=False):
+def run(gui=False, plot=True, n_episodes=2, n_steps=None, save_data=False):
```
adamhall, line 18: Why is the default for `n_episodes` changed? Seems to be done for multiple controllers.

adamhall (commented):
Looks good to me. I'm assuming `mpc_experiment.py` and `lqr_experiment.py` give the same results?

examples/lqr/lqr_experiment.py
Federico-PizarroBejarano: Yeah this is something I was wondering. Ideally all the `_experiment.py` files would share a relatively similar interface and defaults. I added plot because I find it useful to have `gui` and `plot` separated, and I added `n_episodes=2` because I find it useful to test that we can still run multiple episodes and properly get the results and plots, as well as show off how easy it is to run multiple episodes. However, this is really a matter of preference. Happy to revert if you prefer to have another interface and defaults.

Federico-PizarroBejarano:
@adamhall I checked the results. The experiments return different results since we generally changed the parameters and how things are set up, but in general everything is the same. However, I am unsure if this causes errors in iLQR. I elaborated my point in the below comment

examples/lqr/lqr_experiment.py
aghezz1: I like the separation for `gui` and `plot` but i would keep `n_episodes=1`

Federico-PizarroBejarano: reverted to `n_episodes=1`

adamhall:
> @adamhall I checked the results. The experiments return different results since we generally changed the parameters and how things are set up, but in general everything is the same.

Okay cool. All we want from the examples is that they run and show how to set up the controllers. I'm okay to merge.

PR #182, opened 2024-12-04
Improve acados mpc formulation

aghezz1:
Improve the acados MPC formulation on top of #181 
This PR substitutes #175 

**TODO**
- Once #181 is merged, test mpc with casadi/ipopt and mpc with acados. Fully converged solution should be equal (not granted as they might converge to different local minima but in general it happens).


aghezz1:
Moved to #183

PR #185, opened 2025-01-22
Removing info_in_reset

Federico-PizarroBejarano:
`info_in_reset` is a useless configuration flag as it is always set to True. It is no longer used in Gymnasium, which was the original reason for including it. When it is set to False, it causes issues throughout the gym as we have generally programmed the gym always assuming it is True. I have removed it everywhere.


adamhall (approved):
Looks good to me! Just for history tracing, there used to be an issue with running MPC with the info dict, but that was changed a while ago (I double checked) so should be good to go.

Issue #186, opened 2025-02-02
How to visualize in real-time


ssaigarimella:
Hi, I am currently using the Cartpole Stabilization with LQR example. I set the gui=True argument within the run() function in lqr_experiment.py, So I am able to visualize the cartpole stabilization via the PyBullet window:

![Image](https://github.com/user-attachments/assets/db2ff690-f2e0-4421-801d-9ae7a6dadb3d)

However, I am unable to visualize it in real-time. The simulation run quickly moves the cartpole but I want to see this in "wall clock time" in my PyBullet window. How can I do this?

Thanks for the great work!

Federico-PizarroBejarano:
Hello @ssaigarimella, I have been meaning to find a good solution to this, but a simple solution for now is for you to put a `time.sleep(x)` command in the `select_action(...)` function of LQR (in `lqr.py`). Simply slow down the execution such that it is real time, maxing `x` be equal to the inverse of the control frequency. I am certain there is a better solution, I'll try to look into it on Monday, but for now this works.

ssaigarimella:
@Federico-PizarroBejarano the `time.sleep(x) `works perfectly, thanks!

PR #187, opened 2025-02-03
Real-time GUI Visualization

Federico-PizarroBejarano:
Up until now, when the GUI was turned on it ran at the pace of the gym, which is often many times faster than real-time. This caused the visualization to be useless, since you can't actually tell whats happening. This came up recently in https://github.com/utiasDSL/safe-control-gym/issues/186. I have fixed this by simply sleeping the base_experiment when GUI is on. 

I don't foresee any issues, especially since the GUI is used so rarely, unless someone wants the GUI to run quickly to collect a video or something? But I don't expect that would be done for more than 1 or 2 videos since it would take up massive space, so I think it should be ok.

### DIFF
--- a/safe_control_gym/experiments/base_experiment.py
+++ b/safe_control_gym/experiments/base_experiment.py
@@ -48,7 +48,15 @@ class BaseExperiment:
 
         self.reset()
 
-    def run_evaluation(self, training=False, n_episodes=None, n_steps=None, done_on_max_steps=None, log_freq=None, verbose=True, **kwargs):
+    def run_evaluation(self,
+                       training=False,
+                       n_episodes=None,
+                       n_steps=None,
+                       done_on_max_steps=None,
+                       log_freq=None,
+                       verbose=True,
+                       visualization_time_multiplier=1,
+                       **kwargs):
         '''Evaluate a trained controller.
 
         Args:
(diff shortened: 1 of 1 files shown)


safe_control_gym/experiments/base_experiment.py
```diff
@@ -174,6 +174,12 @@ def _select_action(self, obs, info):
             if success:
                 action = self.env.normalize_action(certified_action)
 
+        if self.last_step_timestep is not None and self.env.GUI is True:
+            # Sleep to maintain real-time pacing
+            elapsed = time.time() - self.last_step_timestep
+            time.sleep(max(0, 1.0 / self.env.CTRL_FREQ - elapsed))
```
adamhall, line 180: I'm wondering if we want a parameter to specify the sleep time instead of assuming "realtime" is favourable. I think the goal is to be able to visualize the system when the controller is too fast? It could have the default of being realtime but I think a parameter would be nice?

adamhall (changes requested):

safe_control_gym/experiments/base_experiment.py
Federico-PizarroBejarano: Done!

adamhall (approved):
Looks great!

PR #190, opened 2025-03-03
Minor changes to examples

Federico-PizarroBejarano:
Was responding to an issue and reviewed the examples and found some minor issues:

1. I don't know why the training parameter is set to true in the MPC example, this seems incorrect. 
2. The PID example uses the ctrl.env as the actual environment. This is not necessarily wrong but not aligned with the rest of our examples. 
3. iLQR does not have a reset_before_run function. This causes issues since we switched to traj_step instead of using info in select_action (should at some point align on that). 
4. Removed what seems to be an unnecessary reset call in base_experiment. We generally reset very frequently. 

Let me know if you foresee any issues with these updates.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,7 +26,7 @@ repos:
     -   id: double-quote-string-fixer
 
 -   repo: https://github.com/pycqa/isort
-    rev: 5.13.2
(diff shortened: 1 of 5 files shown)


adamhall (commented):
LGTM

PR #196, opened 2025-05-02
MPC Linting and Documentation

Federico-PizarroBejarano:
Went through and linted all the MPC files (except for `mpc_acados` which is being reworked). Additionally updated the documentation for all files, especially for the GP files. I did that documentation with with Cursor AI so there may be some mistakes, but significantly more documentation than before.

### DIFF
--- a/safe_control_gym/controllers/mpc/linear_mpc.py
+++ b/safe_control_gym/controllers/mpc/linear_mpc.py
@@ -35,9 +35,6 @@ class LinearMPC(MPC):
             terminate_run_on_done=True,
             constraint_tol: float = 1e-8,
             solver: str = 'sqpmethod',
-            # runner args
-            # shared/base args
-            output_dir='results/temp',
             additional_constraints=None,
             use_lqr_gain_and_terminal_cost: bool = False,
             compute_initial_guess_method=None,
@@ -54,7 +51,6 @@ class LinearMPC(MPC):
             terminate_run_on_done (bool): Terminate the run when the environment returns done or not.
             constraint_tol (float): Tolerance to add the the constraint as sometimes solvers are not exact.
             solver (str): Specify which solver you wish to use (qrqp, qpoases, ipopt, sqpmethod)
-            output_dir (str): Output directory to write logs and results.
             additional_constraints (list): List of constraints.
             use_lqr_gain_and_terminal_cost (bool): Use LQR ancillary gain and terminal cost in the MPC.
             compute_initial_guess_method (str): Method to compute the initial guess for the MPC. Options: None, 'ipopt', 'lqr'.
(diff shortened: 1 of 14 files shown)


safe_control_gym/controllers/mpc/linear_mpc.py
```diff
@@ -35,27 +35,23 @@ def __init__(
             terminate_run_on_done=True,
             constraint_tol: float = 1e-8,
             solver: str = 'sqpmethod',
-            # runner args
-            # shared/base args
-            output_dir='results/temp',
```
MingxuanChe, line 40: Is there any reason why the output dir is removed?

safe_control_gym/controllers/mpc/mpc.py
```diff
@@ -29,45 +29,37 @@ def __init__(
             soft_penalty: float = 10000,
             terminate_run_on_done: bool = True,
             constraint_tol: float = 1e-6,
-            # runner args
-            # shared/base args
-            output_dir: str = 'results/temp',
```
MingxuanChe, line 34: The same here.

MingxuanChe (commented):
Overall looks good IMO. But one thing I noticed is that the output_dir of MPC is removed. I would like to know the reason.

safe_control_gym/controllers/mpc/linear_mpc.py
Federico-PizarroBejarano: I removed the output_dir bc it is handled in the kwargs and set in base_controller. I guess we can discuss what we think is better, but I think it is cleaner to only show the args relevant to the controller in the list, and let the base controller args (seed, output_dir, use_gpu, etc) be left in the kwargs. This shouldn't change functionality at all, just how easy it is to read the code

MingxuanChe: I see. I agree that this looks cleaner. But do we have a place to look up what kwargs are available?

MingxuanChe:
LGTM

adamhall (commented):
Changes look good. Do we have confirmation that it runs the same (with the minor changes made)?

safe_control_gym/controllers/mpc/linear_mpc.py
Federico-PizarroBejarano: The kwargs I removed are the one's in base_controller. Unsure what the best way to showcase the possible args while still making the code pretty. Could add a comment beside kwargs referring to the base_controller args? what do you think? Do you think it needs more clarity?

Federico-PizarroBejarano:
> Changes look good. Do we have confirmation that it runs the same (with the minor changes made)?

I just confirmed that mpc, linear_mpc, and mpc_acados work the same in main and this branch for 3D quad tracking and cartpole stabilization. I think there should be no changes to the functionality.

safe_control_gym/controllers/mpc/linear_mpc.py
MingxuanChe: Adding a comment beside kwargs sounds good to me.

PR #197, opened 2025-05-05
General refactoring of MPC code and especially mpc_acados

Federico-PizarroBejarano:
Made some major changes:
- Removed initial guesses entirely
- Removed the timing code
- Moved the `run` function to gp-mpc since it is the only one that uses it. @MingxuanChe said it is also possible to replace this with the base_experiment class, which would be ideal. 
- Improved the reset/reset_before_run code
- Did a major cleanup of mpc_acados, simplifying it as much as possible while including some suggestions from Andrea Ghezzi to prepare for his PR.

### DIFF
--- a/safe_control_gym/controllers/mpc/mpc.py
+++ b/safe_control_gym/controllers/mpc/mpc.py
@@ -28,7 +28,6 @@ class MPC(BaseController):
             soft_penalty: float = 10000,
             constraint_tol: float = 1e-6,
             use_lqr_gain_and_terminal_cost: bool = False,
-            init_solver: str = 'ipopt',
             solver: str = 'ipopt',
             additional_constraints: list = None,
             **kwargs  # Additional args from base_controller.py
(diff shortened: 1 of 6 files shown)


safe_control_gym/controllers/mpc/mpc.py
```diff
@@ -27,9 +26,7 @@ def __init__(
             warmstart: bool = True,
             soft_constraints: bool = False,
             soft_penalty: float = 10000,
-            terminate_run_on_done: bool = True,
             constraint_tol: float = 1e-6,
-            compute_initial_guess_method: str = 'ipopt',
             use_lqr_gain_and_terminal_cost: bool = False,
             init_solver: str = 'ipopt',
```
MingxuanChe, line 31: I realized this init_solver was used for compute_init_guess and now it can be removed

MingxuanChe:
I just find that the variance when running multiple episodes comes from the `self.traj_step = 0` is not called in `reset_before_run`.

safe_control_gym/controllers/mpc/mpc.py
Federico-PizarroBejarano: Fixed!

MingxuanChe:
LGTM

Issue #174, opened 2024-11-05
Passing current trajectory step to `select_action` function


Federico-PizarroBejarano:
In some controllers we are getting the current step of the trajectory from the environment, via the info, such as https://github.com/utiasDSL/safe-control-gym/blob/0d91ee259e66f1234a460afd060e2bb6e94bbb51/safe_control_gym/controllers/pid/pid.py#L94 In others, we keep track of the step in the controller itself via `self.traj_step`, such as https://github.com/utiasDSL/safe-control-gym/blob/0d91ee259e66f1234a460afd060e2bb6e94bbb51/safe_control_gym/controllers/mpc/mpc_acados.py#L337 We should choose one and standardize it. 

The benefit of the info approach is that the iteration is coming from the environment, and thus the controller is stationary, i.e., if you call a select_action function twice with the same arguments you will get the same action. The drawback is that in real experiments on real hardware, there is no `info` dictionary since the environment isn't giving you more information. One way around this is creating an `info` dictionary in the real experiment script and passing that in with the relevant data.

PR #199, opened 2025-05-07
Removing `traj_step`

Federico-PizarroBejarano:
Responding to this issue https://github.com/utiasDSL/safe-control-gym/issues/174 by removing the `traj_step` code and using the `extract_step` function everywhere.


adamhall (approved):
LGTM

PR #201, opened 2025-05-12
acados MPC edits

aghezz1:
I changed the constraints handling for acados, only bounds are supported otherwise I throw an error.
I also scaled the cost matrix since acados wants the continuous time version (they will be automatically integrated in the cost).

Questions for @Federico-PizarroBejarano
- I would be happy to rename "warmstart" into "initial guess"
- Should I introduce different "initial guess" strategies like "ipopt", "lqr", "tracking_cost"? (similar to [here](https://github.com/aghezz1/pept/blob/main/safe_control_gym/controllers/mpc/mpc_acados.yaml)

### DIFF
--- a/safe_control_gym/controllers/mpc/mpc_acados.py
+++ b/safe_control_gym/controllers/mpc/mpc_acados.py
@@ -105,7 +105,7 @@ class MPC_ACADOS(MPC):
         # Set cost (NOTE: safe-control-gym uses quadratic cost)
         ocp.cost.cost_type = 'LINEAR_LS'
         ocp.cost.cost_type_e = 'LINEAR_LS'
-        ocp.cost.W = scipy.linalg.block_diag(self.Q, self.R)
+        ocp.cost.W = scipy.linalg.block_diag(self.Q/self.dt, self.R/self.dt)
         ocp.cost.W_e = self.Q if not self.use_lqr_gain_and_terminal_cost else self.P
         ocp.cost.Vx = np.zeros((ny, nx))
         ocp.cost.Vx[:nx, :nx] = np.eye(nx)
--- a/safe_control_gym/controllers/mpc/mpc_acados.py
+++ b/safe_control_gym/controllers/mpc/mpc_acados.py
@@ -119,42 +119,51 @@ class MPC_ACADOS(MPC):
 
         # Constraints
         # General constraint expressions
-        state_constraint_expr_list = []
-        input_constraint_expr_list = []
         for state_constraint in self.state_constraints_sym:
-            state_constraint_expr_list.append(state_constraint(ocp.model.x))
+            grad_cns_x = cs.Function("grad_cns_x", [ocp.model.x], [cs.jacobian(state_constraint(ocp.model.x), ocp.model.x)])
+            if np.all(grad_cns_x(np.ones(nx)).full() == np.vstack([-np.eye(nx), np.eye(nx)])):
+                cns_x = cs.Function("cns_x", [ocp.model.x], [state_constraint(ocp.model.x)])
+                bounds = cns_x(np.zeros(nx))
+                ocp.constraints.lbx = bounds[:nx].full()
+                ocp.constraints.ubx = -bounds[nx:].full()
+                ocp.constraints.idxbx = np.array([i for i in range(nx)])
+                ocp.constraints.lbx_e = bounds[:nx].full()
+                ocp.constraints.ubx_e = -bounds[nx:].full()
+                ocp.constraints.idxbx_e = np.array([i for i in range(nx)])
+            else:
+                raise ValueError("Constraint type not supported. Support only for bounded constraints expressed as A * x - b <= 0. Check constraints.py.")
         for input_constraint in self.input_constraints_sym:
-            input_constraint_expr_list.append(input_constraint(ocp.model.u))
-
-        h_expr_list = state_constraint_expr_list + input_constraint_expr_list
-        h_expr = cs.vertcat(*h_expr_list)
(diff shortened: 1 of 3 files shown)


safe_control_gym/controllers/mpc/mpc_acados.py
```diff
@@ -119,42 +119,51 @@ def setup_optimizer(self, *args):
 
         # Constraints
         # General constraint expressions
-        state_constraint_expr_list = []
-        input_constraint_expr_list = []
         for state_constraint in self.state_constraints_sym:
-            state_constraint_expr_list.append(state_constraint(ocp.model.x))
+            grad_cns_x = cs.Function("grad_cns_x", [ocp.model.x], [cs.jacobian(state_constraint(ocp.model.x), ocp.model.x)])
```
MingxuanChe, line 123: I think this is a good way to handle the constraint if we let it only take simple constraints.

MingxuanChe: Just curious: I noticed that the bound constraint like `x_lowerbar<= J x` is not available for the initial shooting node [[link to formulation]](https://github.com/acados/acados/blob/main/docs/problem_formulation/problem_formulation_ocp_mex.pdf). Any reason for that?

safe_control_gym/envs/constraints.py
```diff
@@ -178,7 +178,7 @@ def check_tolerance_shape(self):
             raise ValueError('[ERROR] the tolerance dimension does not match the number of constraints.')
 
 
-class QuadraticContstraint(Constraint):
+class QuadraticConstraint(Constraint):
```
MingxuanChe, line 181: Nice catch

MingxuanChe (commented):
Thank you for the PR! I checked the code and it runs well. One minor comment on the possibly missing scaling for the terminal cost. The current way to handle the constraint looks good to me!

safe_control_gym/controllers/mpc/mpc_acados.py
```diff
@@ -119,42 +119,51 @@ def setup_optimizer(self, *args):
 
         # Constraints
         # General constraint expressions
```
MingxuanChe, line 121: Minor comment: This could be removed since we consider only simple bounds.

MingxuanChe:
One additional comment: the current way to handle the constraint doesn't include the bound tolerance `constraint_tol`. IMO, we still keep that, but this could be put in a later PR. Happy to hear the opinions from you guys.

safe_control_gym/controllers/mpc/mpc_acados.py
aghezz1: Which constraints exactly?

safe_control_gym/controllers/mpc/mpc_acados.py
```diff
@@ -105,7 +105,7 @@ def setup_optimizer(self, *args):
         # Set cost (NOTE: safe-control-gym uses quadratic cost)
         ocp.cost.cost_type = 'LINEAR_LS'
         ocp.cost.cost_type_e = 'LINEAR_LS'
-        ocp.cost.W = scipy.linalg.block_diag(self.Q, self.R)
+        ocp.cost.W = scipy.linalg.block_diag(self.Q/self.dt, self.R/self.dt)
         ocp.cost.W_e = self.Q if not self.use_lqr_gain_and_terminal_cost else self.P
```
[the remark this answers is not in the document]
aghezz1: No, because the terminal cost is not integrated by acados and `self.Q` is the cost that you set in the yml file (where you don't make distinction between stage and terminal cost). The scaling of the stage cost is done because acados integrates the stage cost by doing `1/2 * dt * x.T @ Q @ x` with Q from the continuous-time system. Since your `q_mpc` in the yml file is the already the integrated cost (cf. [here](https://github.com/utiasDSL/safe-control-gym/blob/77332383bd5fd2c1715863c19361926c24595fac/safe_control_gym/envs/gym_pybullet_drones/quadrotor.py#L571)) then to obtain the same results in acados we have to do `1/2 * dt * x.T @ (Q/dt) @ x`

safe_control_gym/controllers/mpc/mpc_acados.py
MingxuanChe: I mean eq (7)-(8) type of constraints in the formulation

safe_control_gym/controllers/mpc/mpc_acados.py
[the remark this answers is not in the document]
MingxuanChe: Thank you for the explanation!

Federico-PizarroBejarano (changes requested):

safe_control_gym/controllers/mpc/mpc_acados.py
```diff
@@ -118,43 +118,51 @@ def setup_optimizer(self, *args):
         ocp.cost.yref_e = np.zeros((ny_e, ))
 
         # Constraints
-        # General constraint expressions
-        state_constraint_expr_list = []
-        input_constraint_expr_list = []
         for state_constraint in self.state_constraints_sym:
```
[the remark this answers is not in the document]
aghezz1: Sorry I ran now your code and understood. It works well, I thought you were getting the bounds after the casting into LinearConstraint [here](https://github.com/utiasDSL/safe-control-gym/blob/0b918f54be5c4d5f2b0ac7430b14110b3e8a6b67/safe_control_gym/envs/constraints.py#L323), instead you recover the original bounded constraint.
Good solution 😄 
I can push the update now, together with the linted code

PR #202, opened 2025-05-24
Create per-instance copy of quadrotor randomization attributes

jongoiko:
This small PR makes each `Quadrotor` instance have its own copy of the `INERTIAL_PROP_RAND_INFO` and `INIT_STATE_RAND_INFO` dictionaries, so that the instance-level attributes are modified (depending on the quad type) rather than the class attributes.

The motivation is that I was working on a project where two `Quadrotor` envs are created (one 2D and one 3D), and creating the second env after the first threw a `KeyError`, as does the following example:
```python
from safe_control_gym.envs.gym_pybullet_drones.quadrotor_utils import QuadType
from safe_control_gym.utils.registration import make

env2d = make(
    "quadrotor",
    quad_type=QuadType.TWO_D,
    physics="pyb",
)
_, obs = env2d.reset()

env3d = make(
    "quadrotor",
    quad_type=QuadType.THREE_D,
    physics="pyb",
    task_info=dict(stabilization_goal=[0, 0, 0]),
)
_, obs = env3d.reset()
```

The PR makes the example behave as expected and without errors.

### DIFF
--- a/safe_control_gym/envs/gym_pybullet_drones/quadrotor.py
+++ b/safe_control_gym/envs/gym_pybullet_drones/quadrotor.py
@@ -42,7 +42,7 @@ class Quadrotor(BaseAviary):
         }
     }
 
-    INERTIAL_PROP_RAND_INFO = {
+    BASE_INERTIAL_PROP_RAND_INFO = {
         'M': {  # Nominal: 0.027
             'distrib': 'uniform',
             'low': 0.022,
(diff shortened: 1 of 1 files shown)


Federico-PizarroBejarano:
Hi @jongoiko , thanks for your interest in the gym and the PR! I appreciate the example, made the error very clear. Honestly we never built the gym thinking we would instantiate more than one quadrotor or other robot at a time, I am surprised this is all that it took to get it to work

Federico-PizarroBejarano (approved):
LGTM

PR #205, opened 2025-08-05
Minor updates

Federico-PizarroBejarano:
Updating multiple small issues:

1. Cleaning up `.gitignore` significantly. Unsure why it was so complex to start
2. Improving the way `controller_data` and `safety_filter_data` are logged
3. Cleaning up `mpc_utils.py`
4. Making `ppo_utils.py` a bit more clear
5. Improving the way `constraint_penalty` is set and how it works

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,154 +1,20 @@
-#### Temporary files
-examples/lqr/*data/
-#
-examples/mpsc/temp-data/
-examples/mpsc/unsafe_rl_temp_data/
-#
-examples/pid/*data/
-#
+# Project-specific temporary files
 results/
-z_docstring.py
-TODOs.md
-
-
-
-# macOS users
-.DS_Store
-
-# Byte-compiled / optimized / DLL files
-__pycache__/
-*.py[cod]
-*$py.class
-
-# C extensions
-*.so
-
-# Distribution / packaging
-.Python
-build/
-develop-eggs/
-dist/
-downloads/
-eggs/
-.eggs/
-lib/
-lib64/
-parts/
-sdist/
-var/
-wheels/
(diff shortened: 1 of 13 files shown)


safe_control_gym/envs/benchmark_env.py
```diff
@@ -487,10 +489,10 @@ def after_step(self, obs, rew, done, info):
             if self.constraints is not None and self.use_constraint_penalty and self.constraints.is_violated(self, c_value=c_value):
                 if self.rew_exponential:
                     rew = np.log(rew)
-                    rew += self.constraint_penalty
```
adamhall, line 490: Is this sign change minor? Why are we changing it now?

adamhall (changes requested):
Very small changes. Can ignore if you like.

.gitignore
```diff
@@ -1,154 +1,20 @@
...
-__pypackages__/
-
-# Celery stuff
-celerybeat-schedule
-celerybeat.pid
-
-# SageMath parsed files
-*.sage.py
-
-# Environments
```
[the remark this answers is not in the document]
Federico-PizarroBejarano: Happy to keep useful ones, just wanted to remove bloat. One of the reasons I removed the env and venv stuff is because the gym is indicates using conda, but since this isnt necessary I am happy to add this back in. Will do so now

safe_control_gym/envs/benchmark_env.py
Federico-PizarroBejarano: I switched the sign for this because i found it unintuitive that the _penalty_ is negative. The penalty should be positive, but subtracted from the reward. Pretty minor, don't think most people use this feature

